PR #1, opened 2023-05-09
CLOUDP-175576: repository cleanup from v1 changes

wtrocki:
## Description

> NOTE: you can skip first commit for review 🙈 as it contains only file deletion for v1 SDK

This PR attempts to clean up atlas-sdk go repository from all manual SDK resources. 
We are introducing `core` folder that contains versioning and other common methods that are required by auth.
Please ignore first commit  when reviewing (it does remove all v1 source code)

All changes were introduced according to the spec and scope agreements. 

## Notes

For simplicity of review, I have decided to not rename mongodbatlasv2 folder in this PR. 
Will make rename change in follow-up PRs.

## Documentation

I have updated Contributing Guide covering new repository structure.


auth/oauth.go
```diff
@@ -26,13 +26,13 @@ import (
...
-	atlas "go.mongodb.org/atlas/mongodbatlas"
+	"go.mongodb.org/atlas-sdk/core"
 )
 
-const defaultBaseURL = atlas.CloudURL
+const defaultBaseURL = "https://cloud.mongodb.com/"
 
 var (
-	userAgent = fmt.Sprintf("go-mongodbatlas/%s (%s;%s)", atlas.Version, runtime.GOOS, runtime.GOARCH)
+	userAgent = fmt.Sprintf("go-mongodbatlas/%s (%s;%s)", core.Version, runtime.GOOS, runtime.GOARCH)
```
wtrocki, line 35: We likely need to change this to go-sdk-auth to avoid confusion when we release 0.1.0.

gssbzn (approved):
LGTM

core/client.go
gssbzn, line 1: will the generated SDK use this or is this only for auth?

wtrocki: Yes. Generated SDK does use that at the moment for versioning. I plan to move more internal and non-generated elements to core so we can reuse them. This also enables us to add more autogenerated apis in the future.

wtrocki: Core is the internal implementation of SDK. Users will never need to have imports to the core.

PR #16, opened 2023-05-23
CLOUDP-172196: reflect error when query contains array of objects

wtrocki:
Fixing blocker for CLOUDP-172196 - removing redundant reflect. 

NOTE: openapi generator adds imports for reflection that we cannot remove in the template.

### DIFF
--- a/admin/api_performance_advisor.go
+++ b/admin/api_performance_advisor.go
@@ -487,11 +487,10 @@ func (a *PerformanceAdvisorApiService) listSlowQueriesExecute(r ListSlowQueriesA
 	}
 	if r.namespaces != nil {
 		t := *r.namespaces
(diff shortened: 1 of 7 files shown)


admin/api_performance_advisor.go
```diff
@@ -488,10 +488,7 @@ func (a *PerformanceAdvisorApiService) listSlowQueriesExecute(r ListSlowQueriesA
 	if r.namespaces != nil {
 		t := *r.namespaces
 		if reflect.TypeOf(t).Kind() == reflect.Slice {
```
swattyT, line 490: Do we need this if?

wtrocki: Good point. It is not needed for execution phase. It is because we do import reflect.
I will create workaround for that outside that function

swattyT (approved):
LGTM!

PR #18, opened 2023-05-24
CLOUDP-176386 - format examples for better documentation 

wtrocki:
## Description

We will have a number of examples (we need to test some apis like billing). 
This PR introduces separate README for the examples with instructions for running them.


gssbzn (approved):
LGTM but one thing to consider is using actual go examples as these will be added to the go docs page

wtrocki:
Applying suggestions.

wtrocki:
As for examples format I target end user copying those into some places thus testability would be an disadvantage IMHO

PR #20, opened 2023-05-24
CLOUDP-171748 - Fluent API performance improvements (breaking change)

wtrocki:
## Description

Breaking change performance improvement for the fluent API - passing struct as a pointer.
Improved examples to better cover those details.

Ignore generated changes (dozens of files)


wtrocki:
One of the thoughs I had was to move requestBody to main function instead of calling it from fluent api.
That will assume that request body when present is never optional. If we can assume that I'm happy to do it in follow up PR

gssbzn (approved):
LGTM but it would be good to get @fmenezes opinion as well since he requested this

fmenezes (approved):
thanks for this change

PR #24, opened 2023-05-25
CLOUDP-179041: unify go version for ci/cd

wtrocki:
## Description

CI/CD process is using some old versions of Golang. 
Unifying everything to run on 1.20.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.19
+          go-version: '1.20'
       - name: lint
         uses: golangci/golangci-lint-action@v3.4.0
         with:
@@ -27,7 +27,6 @@ jobs:
     strategy:
       matrix:
         golang:
-          - "1.19"
           - "1.20"
     steps:
       - name: Checkout repository
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,23 +24,12 @@ jobs:
   tests-on-unix:
     needs: golangci-lint # run after golangci-lint action to not produce duplicated errors
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        golang:
-          - "1.20"
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Install Go
         uses: actions/setup-go@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.19
+          go-version: '1.20'
```
wtrocki, line 18: We should be using 1.20 for GA everywhere. 1.19 support is almost guaranteed when we CI/CD against 1.20

fmenezes:
why does our `go.mod` say 1.18 and our CI uses 1.20? is it intended?

.github/workflows/pr.yml
```diff
@@ -15,19 +15,18 @@ jobs:
...
-          version: v1.50.1
+          version: v1.52.2
 
   tests-on-unix:
     needs: golangci-lint # run after golangci-lint action to not produce duplicated errors
     runs-on: ubuntu-latest
     strategy:
       matrix:
         golang:
-          - "1.19"
```
wtrocki, line 30: The only controversial change here. I do not see benefit of us verifying under 2 versions of golang. 
We will likely to introduce another version soon so for GA we should have CI/CD process on single version for simplicty. 

Linting, formatting, introspections are all version based.

fmenezes: should we remove the matrix then?

wtrocki: Good point. We can always restore it for 1.21

.github/workflows/release-check.yml
```diff
@@ -11,7 +10,7 @@ jobs:
     strategy:
       matrix:
         golang:
-          - 1.18
+          - '1.20'
```
fmenezes, line 13: you can remove the step `actions/cache@v3` your `actions/setup-go@v4` already caches it

fmenezes: same for `.github/workflows/pr.yml`

.github/workflows/pr.yml
fmenezes: if we intend to restore it, there is no need to remove

.github/workflows/release-check.yml
wtrocki: Awesome catch! it will make entire pipeline much cleaner.

PR #26, opened 2023-05-25
CLOUDP-171748: fluent API accepting request body in creation method.

wtrocki:
## Description

Adding request body to fluent API:
```
ToggleAWSCustomDNS(ctx context.Context, groupId string)
```
Becomes
```
ToggleAWSCustomDNS(ctx context.Context, groupId string, awsCustomDNSEnabled *AWSCustomDNSEnabled)
```
## Why!

This change was made based on the research study and feedback from CLI migration.
I'm confident that this API will:

1. Make migration from v1 to v2 for other teams much simpler - almost trivial for all "non-GET" requests.
With that change, migration can be automated by regexp search

2. Remove confusion and need to refer to the documentation in order to find how to inject the request body

3. Making the entire API shorter and easier to work with. Developers can make requests for "non-GET" data with single method.

## Downsides

1. Changes in multiple templates for api, docs drive us far from the original OpenAPI generator API?

## Assumptions

1. If no request body presents no argument
2. If the request body is present but optional (nil can be passed)
3. Fluent API for request body is no longer needed (less confusion with optional arguments.

## Rejected ideas

1. Adding request body to  `WithParams` method.
2.  Removing the request body from `WithParams` structure.

## Notes

I removed the authentication-specific code from the template as a bonus.
I have added a dedicated way to generate tests (they aren't committed right now).


fmenezes (approved):
thanks for this

swattyT (approved):
LGTM!

PR #34, opened 2023-05-29
CLOUDP-180035: folder change for toc

wtrocki:
## Description

Changed installation folder for toc tool.

### DIFF
--- a/scripts/toc.sh
+++ b/scripts/toc.sh
@@ -2,16 +2,29 @@
 set -o errexit
 set -o nounset
 
-if which "./scripts/gh-md-toc" >/dev/null 2>&1; then
+BIN_FOLDER=./bin
+DOCS_FOLDER=./docs
+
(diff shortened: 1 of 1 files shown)


scripts/toc.sh
```diff
@@ -5,8 +5,11 @@ set -o nounset
 if which "./scripts/gh-md-toc" >/dev/null 2>&1; then
   echo "Binary gh-md-toc exists."
 else
+  echo Downloading scripts
+  pushd ./scripts
```
gssbzn, line 9: use a different folder, probably `bin` and ignore that folder on git, currently downloading in scripts means it'll be committed

wtrocki: Good suggestion. It is part of the gitignore right now. Using bin folder seems like good standard rather than juggling paths. Making fix.

gssbzn (approved):
LGTM

PR #51, opened 2023-06-01
CLOUDP-173616: remove enuns from schema

wtrocki:
## Description

Removing enums from the schema (thus from the SDK) because they aren't complete and will cause parsing issues with currnet implementation.
Our approach for enums should be to support users by defining enums but not force users to work with predefined objects.


admin/model_alert_config_view_for_nds_group.go
```diff
@@ -15,8 +15,9 @@ type AlertConfigViewForNdsGroup struct {
 	// Date and time when MongoDB Cloud created the alert configuration. This parameter expresses its value in the <a href=\"https://en.wikipedia.org/wiki/ISO_8601\" target=\"_blank\" rel=\"noopener noreferrer\">ISO 8601</a> timestamp format in UTC.
 	Created *time.Time `json:"created,omitempty"`
 	// Flag that indicates whether someone enabled this alert configuration for the specified project.
-	Enabled       *bool                             `json:"enabled,omitempty"`
-	EventTypeName *ServerlessEventTypeViewAlertable `json:"eventTypeName,omitempty"`
+	Enabled *bool `json:"enabled,omitempty"`
+	// Event type that triggers an alert.
+	EventTypeName *string `json:"eventTypeName,omitempty"`
```
matt-condon, line 20: @swattyT here is the fix 😃

matt-condon (approved):
Awesome work, LGTM!

PR #54, opened 2023-06-02
CLOUDP-180579: extra transformations

wtrocki:
## Description

Adding transformations for the OpenAPI files that have been found in the SDK migration
Those are added as a workaround as they will also added to original schema.


tools/transformer/src/atlasTransformations.js
```diff
@@ -95,3 +90,28 @@ module.exports = function runTransformations(openapi) {
 
   return openapi;
 };
+
+function workaroundNestedTransformations(openapi) {
```
wtrocki, line 94: Just extracted function from main file

tools/transformer/src/transformations/oneOf.js
```diff
@@ -90,6 +90,9 @@ function transformOneOfProperties(parentObject, api) {
   );
 
   for (let childObject of childObjects) {
+    if (!childObject.properties) {
```
wtrocki, line 93: This is extra check to ensure corner cases when children do not have any properties.

fmenezes (approved):

PR #63, opened 2023-06-06
CLOUDP-181547: transformation for nested hashmaps

wtrocki:
## Description

New transformation is required to relax json parser for cases when we want to support object and array.
Api with additionalProperties properly generates an map[string]map[string]interface{} objects for SDK.
However in multiple cases for search all we need is  map[string]interface{}

### DIFF
--- a/admin/model_fts_mappings.go
+++ b/admin/model_fts_mappings.go
@@ -14,7 +14,7 @@ type FTSMappings struct {
 	// Flag that indicates whether the index uses dynamic or static mappings. Required if **mappings.fields** is omitted.
 	Dynamic *bool `json:"dynamic,omitempty"`
 	// One or more field specifications for the Atlas Search index. Required if **mappings.dynamic** is omitted or set to **false**.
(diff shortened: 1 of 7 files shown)


admin/model_fts_mappings.go
```diff
@@ -14,7 +14,7 @@ type FTSMappings struct {
 	// Flag that indicates whether the index uses dynamic or static mappings. Required if **mappings.fields** is omitted.
 	Dynamic *bool `json:"dynamic,omitempty"`
 	// One or more field specifications for the Atlas Search index. Required if **mappings.dynamic** is omitted or set to **false**.
-	Fields map[string]map[string]interface{} `json:"fields,omitempty"`
```
wtrocki, line 17: I could easily fix that upstream

wtrocki:
@mongodb/apix Giving some extra context here. We need this transformation as a temporary measure to fix original inconsistencies on our side. This cannot be fixed in OpenAPI generator as OpenAPI has array and object are two separate types in the spec.

PR #67, opened 2023-06-08
CLOUDP-182930:  fix discriminator parser 

wtrocki:
## Description

Initially our design for SDK assumed that if we missing discriminator and we cannot in clear way make API to return one instance we should fail. This is semanitically valid for oneOf statement, but.. it does generate number of failures that are usually found at the later state of using SDK.  

This PR changes our approach for oneOf handling giving users the ability to access multiple parsed objects. 
This requires some validation and discussion but it seems like an alternative to just returning errors.
  
 NOTE: all modified source code in this PR might actually require discriminator mapping upstream or adding transformation extension to openapi. This way we going to solve problems globally.

### DIFF
--- a/admin/model_create_endpoint_request.go
+++ b/admin/model_create_endpoint_request.go
@@ -79,12 +79,9 @@ func (dst *CreateEndpointRequest) UnmarshalJSON(data []byte) error {
 	}
 
 	if match > 1 { // more than 1 match
-		// reset to nil
-		dst.CreateAWSEndpointRequest = nil
-		dst.CreateAzureEndpointRequest = nil
-		dst.CreateGCPEndpointGroupRequest = nil
-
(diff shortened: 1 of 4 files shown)


wtrocki:
Thinking about it I have mixed feelings for introducing that PR. 
We cannot perform effective validation for this as oneOf is perfectly valid without discrimination.

matt-condon:
Main question for me with this PR is the consequences of either decision (keeping or removing the validation)

I think if we keep the validation on the sdk side, we probably need to introduce some additional validation to ensure oneOf schemas have valid discriminators, otherwise there's a risk of introducing further regressions

If we go with this change, we lose sdk validation, at the potential cost of less verbose errors for the user, but there's less of a risk of introducing regressions

wtrocki:
Agree. Closing

PR #71, opened 2023-06-13
Use extension to mark experimental

wtrocki:
## Description

The proposal is to use x-experimental to mark the field as experimental. 
This will allow us to adjust templates in various places to include that information.


matt-condon (approved):
Really nice PR, thank you for this!

PR #75, opened 2023-06-13
CLOUDP-170462: make transformation resilient to renames

wtrocki:
No transformation changes. Provided both names so we can gracefully fail when model is renamed. 
Transformation needs to stay but names will be constant for now.


swattyT (approved):
LGTM!

PR #76, opened 2023-06-14
CLOUDP-170462: rename region config

wtrocki:
## Description

RegionConfig name is too generic. Changed to CloudRegionConfig

### DIFF
--- a/tools/transformer/src/atlasTransformations.js
+++ b/tools/transformer/src/atlasTransformations.js
@@ -107,7 +107,7 @@ module.exports = function runTransformations(openapi) {
 };
 
 function workaroundNestedTransformations(openapi) {
-  var parentObject;
+  let parentObject;
   try {
     parentObject = getObjectFromYamlPath(
       ".components.schemas.RegionConfig",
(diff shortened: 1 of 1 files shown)


tools/transformer/src/atlasTransformations.js
```diff
@@ -107,15 +107,22 @@ module.exports = function runTransformations(openapi) {
 };
 
 function workaroundNestedTransformations(openapi) {
+  var parentObject;
```
gssbzn, line 110:
```suggestion
  let parentObject;
```

tools/transformer/src/atlasTransformations.js
```diff
@@ -107,15 +107,22 @@ module.exports = function runTransformations(openapi) {
...
       ".components.schemas.RegionConfig",
       openapi
     );
-    if (parentObject) {
-      transformOneOfProperties(parentObject, openapi);
-    }
-  } catch (e) {
+  } catch (e) {}
+  try {
+    parentObject = getObjectFromYamlPath(
```
tibulca, line 118:
```suggestion
    parentObject = parentObject || getObjectFromYamlPath(
```

if parentObject is successfully set at line 112, then there's no need to set it here again, right?

tibulca (approved):

PR #80, opened 2023-06-15
fix: remove unneeded reviewers

wtrocki:
We do have codeowners in place for all repo so dependabot will not be able to add those


wtrocki:
Note: need to create jira for that change.

gssbzn (approved):

PR #82, opened 2023-06-15
fix: rename example file to match expected naming

gssbzn:
## Description

Rename file to match expected naming practices for examples


wtrocki:
Awesome catch! Verifying if that doesn't affect our backend builds.

wtrocki (approved):

PR #83, opened 2023-06-15
fix: godoc naming to match the function

gssbzn:
## Description

Fix naming to make godocs happy

### DIFF
--- a/admin/atlas_client.go
+++ b/admin/atlas_client.go
@@ -13,11 +13,11 @@ import (
 )
 
 const (
-	// CloudURL is default base URL for the services.
+	// DefaultCloudURL is default base URL for the services.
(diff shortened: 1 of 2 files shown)


wtrocki:
Need more context? Why godoc? I guess this is for https://pkg.go.dev/go.mongodb.org/atlas-sdk

wtrocki (approved):

admin/atlas_client.go
```diff
@@ -13,11 +13,11 @@ import (
 )
 
 const (
-	// CloudURL is default base URL for the services.
+	// DefaultCloudURL is default base URL for the services.
```
wtrocki, line 16: This will be overrwriten by generator. We need to change template

PR #85, opened 2023-06-15
fix: drop generators as a public package

gssbzn:
## Description

Currently the generators package is publicly documented and could be imported and used by 3rd parties, limiting that with this PR


wtrocki (approved):
LGTM

PR #86, opened 2023-06-15
CLOUDP-184833: temporary transform for the change that generates inva…

wtrocki:
In order to generate dev changes we need to provide this temporary workaround:
I have created separate issue to fix this problem upstream but do not want us to have blocked ability to get dev changes during that time.

https://jira.mongodb.org/browse/CLOUDP-184861


gssbzn (approved):
LGTM

PR #92, opened 2023-06-15
CLOUDP-184889: remove API path from method list

wtrocki:
Community feedback that table is to wide and paths aren't adding any benefits. Individual paths can be also seen in the details if needed


wtrocki:
@sarahsimpers Do you have opinion about this change? Does it make sense?

sarahsimpers:
> @sarahsimpers Do you have opinion about this change? Does it make sense?

This works for me! Thanks

PR #93, opened 2023-06-15
CLOUDP-184833: documentation update

wtrocki:
## Description

PR contains only autogenerated documentation. 
I did verified the documentation separately. It wasn't included in the original backport PR to not obfuscate review.
Normally documentation should be automatically generated in the PR after but since we are in backport mode we cannot merge those PRs effectively.


docs/docs/AvailableRegion.md
drinkbird, line 1: What causes such deletions?

wtrocki: Model rename is Addition and Deletion in the sdk

drinkbird (approved):

PR #103, opened 2023-06-23
CLOUDP-185937: merge integrations

wtrocki:
## Description

If we add or remove future integrations we would not be able to parse those results. 
I have modified oneOf transformation to circumvent main issue we could not merge those (duplicate fields).

We will now fail if field is in different type but also include descriptions


tools/transformer/src/engine/transformers.js
```diff
@@ -15,20 +15,28 @@ function filterObjectProperties(object, filter = (_k, _v) => true) {
...
   const duplicates = [];
   for (const obj of objArray) {
     if (obj) {
       for (const key of Object.keys(obj)) {
-        if (allKeys.has(key)) {
-          duplicates.push(key);
+        if (allKeys[key]) {
+          const parentType = allKeys[key].type;
+          const childType = obj[key].type;
+          const typeMissmatch = parentType? parentType!== childType: false;
```
matt-condon: slight typo `typeMismatch`

matt-condon (approved):

PR #114, opened 2023-06-30
CLOUDP-187123: improve error handling for inline polymorphism.

wtrocki:
## Motivation

When validating schema we noticed that one of the errors in the transformation validation doesn't 
provide enough context. This one-liner helps backend users to understand what is wrong with their schema.

### DIFF
--- a/tools/transformer/src/engine/readers.js
+++ b/tools/transformer/src/engine/readers.js
@@ -118,6 +118,8 @@ function getNameFromYamlPath(path) {
 }
 
 function getObjectFromYamlPath(path, obj) {
+  errMessage = `Your schema contain inline openapi objects. 
+  This usually happens when schema contains oneOf or anyOf statements directly in the path.`;
   const propertiesStack = path.split(".").reverse();
   let currObj = obj;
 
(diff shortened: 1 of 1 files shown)


tools/transformer/src/engine/readers.js
```diff
@@ -118,6 +118,8 @@ function getNameFromYamlPath(path) {
 }
 
 function getObjectFromYamlPath(path, obj) {
+  errMessage = `Your schema contain inline openapi objects.
```
matt-condon, line 121:
```suggestion
  errMessage = `Your schema contains inline openapi objects. 
```

tools/transformer/src/engine/readers.js
```diff
@@ -118,6 +118,8 @@ function getNameFromYamlPath(path) {
 }
 
 function getObjectFromYamlPath(path, obj) {
+  errMessage = `Your schema contain inline openapi objects. 
+  This usually happens when schema contains oneOf or anyOf statements directly in the response or request body.`;
```
matt-condon, line 122: Nit; Worth including a suggestion to mitigate against this specific issue?

matt-condon (approved):

tools/transformer/src/engine/readers.js
wtrocki: I think I have provided an generic open API mitigation. Not sure if that would not confuse backend engineers? looking for feedback. Would you understand what you need to do?

PR #115, opened 2023-07-03
CLOUDP-188102: remove nullability from the SDK

wtrocki:
## Description

All nullable properties aren’t meant to be nullable but ignored when sending request.

TL;DR: Instead of: { “field”: null }  we look for {} 

Additionally nullablity in the SDK is implemented in the way that provides significant usability issues. All nullable objects need to be created as they aren’t pointers and cannot be skipped. 
In all existing cases backend rejects null values thus will make those SDK methods unusable.

Nullability have been covered in[ small subset](https://github.com/mongodb/atlas-sdk-go/pull/115/commits/8b78624094b2d28ea32d76a557d9f51490ae1f11) of the SDK objects with most of them not covered by Atlas CLI. Old SDK did not reflect any nullability


gssbzn (approved):

wtrocki:
Follow up PR will introduce release notes and documentation

PR #116, opened 2023-07-05
CLOUDP-188215: add type info into generated code

wtrocki:
## Description

Based on discussions with various teams - Flattening for the schema is the most efficient solution for now but we need to be able to counteract missing type info in user facing artifacts. Based on the investigation in this PR I see that usefulness of adding types can be questioned. 

In some cases, it adds clarity but in others, it does introduce confusion

### DIFF
--- a/openapi/atlas-api-transformed.yaml
+++ b/openapi/atlas-api-transformed.yaml
@@ -14883,8 +14883,8 @@ components:
           type: string
           format: date-time
           description: >-
-            Date and time until which this alert has been acknowledged. This
-            parameter expresses its value in the <a
+            DefaultAlertViewForNdsGroup: Date and time until which this alert
+            has been acknowledged. This parameter expresses its value in the <a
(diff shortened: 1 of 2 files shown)


openapi/atlas-api-transformed.yaml
```diff
@@ -23469,42 +23580,44 @@ components:
         endHour:
           type: integer
           format: int32
-          description: Hour of the day when the scheduled window to run one online archive
-            ends.
+          description: "MonthlyScheduleView: Hour of the day when the scheduled window to
```
wtrocki, line 23583: Good example of increased clarity.

wtrocki:
CC: @gssbzn I think that based on that investigation we can look to introduce more clarity in descriptions themselves. 
This POC supports objects names which has mixed results. 
I will investigate if we can use discriminator/Type values which should be less internal and introduce clarity.

wtrocki:
Alternative approach: https://github.com/mongodb/atlas-sdk-go/pull/117

PR #120, opened 2023-07-21
CLOUDP-191409: date field best practices

wtrocki:
## Description

Ensuring we have clarity about date handling in the SDK. 
Clarifies how to map date to the string and check for date existence. 

NOTE: advice here is generic golang advice for dealing with the time.Time objects and pointers.

Based on threads in the following PRs:
https://github.com/mongodb/mongodb-atlas-cli/pull/2096
https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1324

### DIFF
--- a/docs/doc_5_best_practices.md
+++ b/docs/doc_5_best_practices.md
@@ -30,16 +30,16 @@ if test != ""
 
 ## Working with Date Fields
 
-In the Atlas SDK GO, date fields are represented by `*time.Time` type to handle time-related data. 
-When comparing values that are `time.Time` based never compare pointers directly or 
+In the Atlas Go SDK, the `*time.Time` type represents date fields for handling time-related data. 
+When you compare values based on `time.Time`, either never compare pointers directly or do the following:
 
-1. Avoid using direct comparison operators (e.g., `myStruct.MyDateField == ""`) to check for equality. Comparing pointers directly will check if they refer to the same memory address rather than comparing the actual date values. Since each time.Time instance is allocated in different memory locations, direct comparisons may yield unexpected results.
+1. Avoid using direct comparison operators (e.g., `myStruct.MyDateField == ""`) to check for equality. Comparing pointers directly will check if they refer to the same memory address rather than comparing the actual date values. Since each `time.Time` instance is allocated in different memory locations, direct comparisons might yield unexpected results.
 
-2. Use the `Has` Function for Checking Non-Nil Pointers:
-The SDK provides a dedicated `HasFieldName` or `GetFieldName` function for each model to check if a time.Time pointer is not nil before accessing its value. Always use this function to ensure that the pointer is valid before performing any operations.
+2. Use the `Has` function to check non-nil pointers:
+The SDK provides a dedicated `HasFieldName` or `GetFieldName` function for each model to check if a `time.Time` pointer is non-nil before accessing its value. Always use this function to ensure that the pointer is valid before you perform any operations.
 
-3. Use time.Time Methods for Comparing Date Values:
-When you have confirmed that the time.Time pointer is not nil, you can safely use `time.Time` methods to compare the actual date values. Commonly used methods for comparison include `Before`, `After`, and `Equal`:
+3. Use `time.Time` methods to compare date values:
+When you have confirmed that the `time.Time pointer` is non-nil, you can safely use `time.Time` methods to compare the actual date values. Commonly used methods for comparison include `Before`, `After`, and `Equal`:
 ```go
     if datePtr1.Before(*datePtr2) {
         // datePtr1 is before datePtr2.
--- a/docs/doc_5_best_practices.md
+++ b/docs/doc_5_best_practices.md
@@ -41,6 +41,10 @@ The SDK provides a dedicated `HasFieldName` or `GetFieldName` function for each 
 3. Use `time.Time` methods to compare date values:
 When you have confirmed that the `time.Time pointer` is non-nil, you can safely use `time.Time` methods to compare the actual date values. Commonly used methods for comparison include `Before`, `After`, and `Equal`:
 ```go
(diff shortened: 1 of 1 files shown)


docs/doc_5_best_practices.md
```diff
@@ -28,6 +28,27 @@ if test != ""
...
 
+## Working with Date Fields
+
+In the Atlas SDK GO, date fields are represented by `*time.Time` type to handle time-related data. 
+When comparing values that are `time.Time` based never compare pointers directly or 
+
+1. Avoid using direct comparison operators (e.g., `myStruct.MyDateField == ""`) to check for equality. Comparing pointers directly will check if they refer to the same memory address rather than comparing the actual date values. Since each time.Time instance is allocated in different memory locations, direct comparisons may yield unexpected results.
+
+2. Use the `Has` Function for Checking Non-Nil Pointers:
+The SDK provides a dedicated `HasFieldName` or `GetFieldName` function for each model to check if a time.Time pointer is not nil before accessing its value. Always use this function to ensure that the pointer is valid before performing any operations.
```
VincentPikand, line 39: "code blockize" code snippets, even if they are a single word

```suggestion
The SDK provides a dedicated `HasFieldName` or `GetFieldName` function for each model to check if a `time.Time` pointer is not `nil` before accessing its value. Always use this function to ensure that the pointer is valid before performing any operations.
```

docs/doc_5_best_practices.md
```diff
@@ -28,6 +28,27 @@ if test != ""
...
+
+3. Use time.Time Methods for Comparing Date Values:
+When you have confirmed that the time.Time pointer is not nil, you can safely use `time.Time` methods to compare the actual date values. Commonly used methods for comparison include `Before`, `After`, and `Equal`:
```
VincentPikand, line 42: same as before

```suggestion
When you have confirmed that the `time.Time` pointer is not `nil`, you can safely use `time.Time` methods to compare the actual date values. Commonly used methods for comparison include `Before`, `After`, and `Equal`:
```

docs/doc_5_best_practices.md
VincentPikand, line 39: Kind of an unrelated question, but aren't the `hasFieldName` and `getFieldName` methods kind of pointless? Like, what's the difference between getting a `false` on `hasFieldName` vs getting `nil` when trying to access it to be sure of it's value? With `getFieldName` the difference is even smaller.

docs/doc_5_best_practices.md
```diff
@@ -28,6 +28,27 @@ if test != ""
...
+```go
+    if datePtr1.Before(*datePtr2) {
+        // datePtr1 is before datePtr2.
+    } else if datePtr1.After(*datePtr2) {
+        // datePtr1 is after datePtr2.
+    } else {
+        // datePtr1 and datePtr2 are equal.
+    }
```
VincentPikand, line 50: this example is very nice, but for consistency with your second point maybe it could be extended to include a `hasFieldName` call

docs/doc_5_best_practices.md
corryroot:
```suggestion
In the Atlas GO SDK, the `*time.Time` type represents date fields for handling time-related data. 
When you comparing values based on `time.Time`, either never compare pointers directly or do the following:

1. Avoid using direct comparison operators (e.g., `myStruct.MyDateField == ""`) to check for equality. Comparing pointers directly will check if they refer to the same memory address rather than comparing the actual date values. Since each `time.Time` instance is allocated in different memory locations, direct comparisons might yield unexpected results.

2. Use the `Has` function to check non-nil pointers:
The SDK provides a dedicated `HasFieldName` or `GetFieldName` function for each model to check if a `time.Time` pointer is non-nil before accessing its value. Always use this function to ensure that the pointer is valid before you perform any operations.

3. Use time.Time Methods for Comparing Date Values:
When you have confirmed that the `time.Time pointer` is non-nil, you can safely use `time.Time` methods to compare the actual date values. Commonly used methods for comparison include `Before`, `After`, and `Equal`:
```

corryroot (commented):
I've commented with minor copy suggestions.

docs/doc_5_best_practices.md
wtrocki: Very good question.

> getFieldName pointless 

Golang requires boilerplate code for the unified handling of the pointer values. 
We strive for better UX by giving developers a unified way to access fields.
Example:
https://github.com/mongodb/atlas-sdk-go/blob/168798f1930c1fbc2484973f23a4fc904e2adec5/admin/model_dedicated_hardware_spec.go#L43-L49

We also provide direct structure access for advanced users. 

> difference between getting a false on hasFieldName vs getting nil

`if struct.value == nil` is not reflecting every case for nullability in golang. 
SDK take on nullability:
https://github.com/mongodb/atlas-sdk-go/blob/168798f1930c1fbc2484973f23a4fc904e2adec5/admin/utils.go#L39

We could improve it over time without affecting users.

docs/doc_5_best_practices.md
wtrocki:
```suggestion
   if !sdkModel.HasSomeDateField() {
       return;
   }
    datePtr1 := sdkModel.SomeDateField;
    if datePtr1.Before(*datePtr2) {
        // datePtr1 is before datePtr2.
    } else if datePtr1.After(*datePtr2) {
        // datePtr1 is after datePtr2.
    } else {
        // datePtr1 and datePtr2 are equal.
    }
```

docs/doc_5_best_practices.md
wtrocki: Since that implementation was never written by us (it came from the upstream) I think we can simplify it to some extent and maybe avoid reflection. Will create separate jira

corryroot (approved):
LGTM!

PR #124, opened 2023-07-26
fix: remove docs from readme.md change as it is autogenerated

wtrocki:
We strive to ping docs team only for manual changes and have reviewer assignment actionable.


andreaangiolillo (approved):
LGTM

PR #129, opened 2023-07-26
Draft Release automation

wtrocki:
## Description

This PR creates release notes and draft releases in github. 
We will start with a draft release first and then have separate PR to set the release's property to false once we have the rest of the automation in place.

Follow-up PRs will introduce multi-version support, s3 automation etc.

## Verification

CI/CD runs with success. The release is created with breaking changes included.
https://github.com/mongodb/atlas-sdk-go/actions/runs/5673051560/job/15373647341?pr=129


.github/workflows/autorelease.yaml
```diff
@@ -0,0 +1,24 @@
+name: Autorelease
+on:
+  push:
+    branches:
+      - main
+      - release-automation
```
wtrocki, line 6: Will be removed after review. This is only to test job. 
Job runs right now and created valid release.
```suggestion
```

BREAKING_CHANGES.md
```diff
@@ -1,20 +0,0 @@
-## v20230202003
```
wtrocki, line 1: We split this file into changes per release for clarity.

tools/releaser/breaking_changes/v20230201002.md
```diff
@@ -0,0 +1,15 @@
+## Breaking Changes
```
andreaangiolillo, line 1: How is this file generated for a new release? Is this a manual process?

.github/workflows/autorelease.yaml
andreaangiolillo, line 6: what changes are inside in release-automation that are not in the main branch?

.github/workflows/autorelease.yaml
```diff
@@ -0,0 +1,24 @@
+name: Autorelease
+on:
```
andreaangiolillo, line 2: I would add   `workflow_dispatch: {}` to be able to run the action manually -> easier to test so you don't have to merge a commit into the branch (feel free to ignore)
```suggestion
on:
  workflow_dispatch: {} # workflow can be run manually
```

andreaangiolillo (approved):
LGTM

.github/workflows/autorelease.yaml
wtrocki: This is only to enable verification of the PR. Removing now

.github/workflows/autorelease.yaml
wtrocki: Amazing point. Will add that as well but it needs some docs

tools/releaser/breaking_changes/v20230201002.md
wtrocki: Yes. Establishing breaking changes is the only manual process. 
It is based on exemptions in openapi.

PR #131, opened 2023-07-27
CLOUDP-155909: breaking changes support plan

wtrocki:
## Description

Adding extra notes covering breaking changes that aren't part of the version. 
In long run occurence of those should be rare but for now we need to define an process for handling those


matt-condon (approved):
LGTM

PR #172, opened 2023-08-23
CLOUDP-195429: detect breaking changes in the SDK

wtrocki:
## Description

Detect breaking changes in the SDK using separate optional jobs.
Job failing will mean that we have breaking changes in place and need to take action by documenting changes.

## How this works

1. Job always compares with the main branch
2. It lists all the incompatibilities (it will be subject to false positives which is fine)
3. It is introduced as a separate step so it will be optional and only act as a helper for developers to verify breaking changes

## Not covered 

A lot of improvements in the upstream job is needed to be able to use this action for automatic merges.
I will be looking to help improve that job so we can rely on it with our current automation and minimize false positives, duplicates etc.

## Job in "action"

https://github.com/mongodb/atlas-sdk-go/actions/runs/5950387722/job/16138071247

### DIFF
--- a/.github/workflows/breaking-changes.yaml
+++ b/.github/workflows/breaking-changes.yaml
@@ -10,5 +10,5 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: '1.20'
+        go-version-file: 'go.mod'
     - uses: joelanford/go-apidiff@main
--- a/.github/workflows/breaking-changes.yaml
+++ b/.github/workflows/breaking-changes.yaml
@@ -1,4 +1,4 @@
-name: Detect Breaking Changes
+name: (Optional) Detect Breaking Changes
 on: [ pull_request ]
 jobs:
   go-apidiff:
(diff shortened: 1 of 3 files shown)


.github/workflows/breaking-changes.yaml
```diff
@@ -0,0 +1,14 @@
...
+  go-apidiff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Install Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.20'
```
gssbzn, line 13: can you set this to use the one from the go.mod?

wtrocki: Our CI/CD is using 1.20: https://github.com/mongodb/atlas-sdk-go/blob/main/.github/workflows/pr.yml#L18

There is a big gap in golang AST between those versions so Golangcilint and other extensions can work more efficiently. There is no impact on using the latest go for this job with actual lowest version of the 

[Offtopic] We could consider bumping version as atlas-cli and terraform seems to use 1.20:

https://github.com/mongodb/mongodb-atlas-cli/blob/master/go.mod#L3

gssbzn (approved):
LGTM

PR #179, opened 2023-08-29
CLOUDP-197445: paths to the version.go file

wtrocki:
## Description

Paths should be git repository paths from git diff.

`./` is not valid prefix as paths aren't executed by bash - they are matched against git diff.


gssbzn (approved):

andreaangiolillo (approved):
LGTM

PR #193, opened 2023-09-13
fix: add missing line break to breaking changes file

wtrocki:
## Description

 When creating the release I noticed that breaking changes were added with a missing line breaks. 
 This small single-line fix will add the required line break to enable us to automatically release.

### DIFF
--- a/tools/releaser/scripts/update-version.sh
+++ b/tools/releaser/scripts/update-version.sh
@@ -19,7 +19,7 @@ echo "Modifying $SDK_MAJOR_VERSION to $BUMPED_MAJOR_VERSION Resource Version acr
 npm exec -c "replace-in-file /$SDK_MAJOR_VERSION/g $BUMPED_MAJOR_VERSION $VERSION_UPDATE_PATHS --isRegex"
 
 echo "Creating empty breaking changes file for $BUMPED_MAJOR_VERSION"
-echo "\n# Breaking Changes" > "$script_path/../breaking_changes/${BUMPED_MAJOR_VERSION}.md"
+printf "\n# Breaking Changes" > "$script_path/../breaking_changes/${BUMPED_MAJOR_VERSION}.md"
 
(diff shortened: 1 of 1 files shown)


tools/releaser/scripts/update-version.sh
```diff
@@ -19,7 +19,7 @@ echo "Modifying $SDK_MAJOR_VERSION to $BUMPED_MAJOR_VERSION Resource Version acr
 npm exec -c "replace-in-file /$SDK_MAJOR_VERSION/g $BUMPED_MAJOR_VERSION $VERSION_UPDATE_PATHS --isRegex"
 
 echo "Creating empty breaking changes file for $BUMPED_MAJOR_VERSION"
-echo "# Breaking Changes" > "$script_path/../breaking_changes/${BUMPED_MAJOR_VERSION}.md"
+printf "\n# Breaking Changes\n" > "$script_path/../breaking_changes/${BUMPED_MAJOR_VERSION}.md"
```
gssbzn, line 22:
```suggestion
echo "# Breaking Changes\n" > "$script_path/../breaking_changes/${BUMPED_MAJOR_VERSION}.md"
```

wtrocki: shellcheck suggests printf :(

gssbzn: oh well, can you then just remove the opening `\n` I don't think is necessary

wtrocki: https://github.com/mongodb/atlas-sdk-go/releases/tag/v20230201008.0.0 seems to be missing `\n` at the beginning for some reason

gssbzn: that seems to be a problem of the releaser not of this script

wtrocki: Closing PR then. Sorry for confusion

PR #200, opened 2023-09-18
fix: manual update of openapi generator

wtrocki:
## Description

Automatic updates thru dependabot would not support openapi generator. 
This PR attempts to verify how many breaking changes we would get when updating openapi generator to latest.


admin/api_access_tracking.go
```diff
@@ -11,7 +11,7 @@ import (
 	"strings"
 )
 
-type AccessTrackingApi interface {
+type AccessTrackingAPI interface {
```
wtrocki, line 14: A pretty noticeable change. 
7.0  renames Api => API. It will break every possible SDK method call and example.

wtrocki: Rename happened in java codebase making it hard to adjust change to become non breaking.

PR #213, opened 2023-10-04
Improve PR description of the automatic PR

wtrocki:
## Description

Add note about the need to reopen PR to trigger builds.

### DIFF
--- a/.github/workflows/autoupdate-prod.yaml
+++ b/.github/workflows/autoupdate-prod.yaml
@@ -53,23 +53,25 @@ jobs:
             2. Documentation style
             3. Transformation engine linting
 
+            > NOTE: New automation PRs need to be closed and reopened to trigger builds
+
             ## Important Details
 
-            1. If build is failing please check the logs and fix the issues in the OpenAPI file (openapi/atlas-api.yaml)
-            2. Do not attempt to manually fix any issues in the Golang Codebase as all changes are automatically generated. 
-            Fixes should be applied to the OpenAPI file (openapi/atlas-api.yaml) or ./tools engine in separate PR.
+            1. If the build is failing please check the logs and fix the issues in the OpenAPI file (openapi/atlas-api.yaml)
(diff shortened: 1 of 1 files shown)


.github/workflows/autoupdate-prod.yaml
```diff
@@ -53,23 +53,25 @@ jobs:
...
+            1. If the build is failing please check the logs and fix the issues in the OpenAPI file (openapi/atlas-api.yaml)
+            2. Do not attempt to manually fix any issues in the Golang code as all changes are automatically generated. 
+            Fixes should be applied to the OpenAPI file (openapi/atlas-api.yaml) or ./tools engine in a separate PR.
 
             ## Manual Review Procedure
 
             1. Review changes in the OpenAPI file (openapi/atlas-api.yaml)
-            2. If we see any breaking changes without new resource version (
-              - Review breaking changes listed in `Detect Breaking Changes` action
+            2. If we see any breaking changes without a new resource version.
```
andreaangiolillo, line 67:
```suggestion
            2. If we see any breaking changes without a new resource version.
```

What does "without a new version" mean here? Could you provide an example?

wtrocki: https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_1_concepts.md#example-version-v2023020100100

wtrocki:
```suggestion
            2. If we see any breaking changes [without a new resource version](https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_1_concepts.md#example-version-v2023020100100) .
```

andreaangiolillo: mh.. sorry but it is still not clear 😭 I am wondering if we should look for a specific property in the code or something else. What you shared is only the versioning strategy of the SDK.

andreaangiolillo: Thanks for the clarification!

andreaangiolillo (approved):
LGTM

PR #214, opened 2023-10-04
CLOUDP-204045:  Fully automatic release procedure

wtrocki:
## Description

I'm fixing bug causing build failures for Breaking changes job.
Fix introduces Autogeneration of Breaking Changes to introduce a fully automatic release procedure without the need to edit the breaking changes file.

## Why

Breaking changes job failures blocking merges of the PR. 
Failures are almost always false positives.
The review process requires users to understand all possible versioned API use cases.

## Issue
 
Job ran after major version bumps causing import incompatibilities.
We were unable to provide breaking changes for the SDK.
We needed to run breaking changes before the bump to not introduce additional import changes.

## Impact

After this change, we no longer need any manual changes during review. All changes happen automatically.

## Verification

Verified flows by running scripts locally with different input in version.json files.


.github/workflows/breaking-changes.yaml
```diff
@@ -1,24 +0,0 @@
-name: (Optional) Detect Breaking Changes
```
wtrocki, line 1: This job needs to run at specific time and cannot be it's own workflow

.github/workflows/autoupdate-prod.yaml
```diff
@@ -64,14 +66,9 @@ jobs:
             ## Manual Review Procedure
 
             1. Review changes in the OpenAPI file (openapi/atlas-api.yaml)
-            2. If we see any breaking changes [without a new resource version](https://github.com/mongodb/atlas-sdk-go/tree/main/tools/releaser#new-resource-version).
```
wtrocki, line 67: Simplification for the reviewers. We no longer need to understand Versioned API and SDK versioning scheme to review PR.

tools/releaser/README.md
```diff
@@ -19,13 +19,6 @@ In situations where we do not want to perform automatic release please remove ch
 
 Manual release is possible by updating `./internal/core/version.go` file with minor version bump.
 
-## Manual Major Release
```
wtrocki, line 22: No need for manual major releases. All will happen automatically.

tools/releaser/README.md
```diff
@@ -37,17 +30,20 @@ For the new Resource Version introduced in the versions.json
...
-### New Generation
+### New SDK Generation without change in Rersource version
 
 - Issue new Minor release
 - Update versions in versions.go
 
-### Manual Breaking Changes Release
+### Breaking Changes Release
+
+In situations when breaking changes are present automation will recognize them and automatically create major version bump.
```
wtrocki, line 42: No need to automatically run the job

tools/releaser/scripts/new-version.sh
```diff
@@ -17,16 +17,29 @@ if [ "$NEW_RESOURCE_VERSION" == "$SDK_RESOURCE_VERSION" ]; then
 	new_minor_version=$((SDK_MINOR_VERSION + 1))
 	# Update the SDK_VERSION
 	SDK_VERSION="${SDK_MAJOR_VERSION}.${new_minor_version}.0"
+
+	echo "Print breaking changes"	
+	# shellcheck source=/dev/null
+	source "$script_path/breaking-changes.sh"
+	if [ -n "$BREAKING_CHANGES" ]; then
```
wtrocki, line 24: This weren't previously automated.

tools/releaser/README.md
```diff
@@ -37,17 +30,20 @@ For the new Resource Version introduced in the versions.json
 - Reset the minor version
 - Replace all occurrences of the Major version in the repository
 - Update versions in versions.go
+- Generate breaking changes file
 
-### New Generation
+### New SDK Generation without change in Rersource version
```
AgustinBettati, line 35:
```suggestion
### New SDK Generation without change in Resource version
```

.github/workflows/autoupdate-prod.yaml
matt-condon: QQ, does this mean we're comfortable assuming all SDK breaking changes will be caught in the generated file (no need to double check)?

tools/releaser/scripts/new-version.sh
```diff
@@ -17,16 +17,29 @@ if [ "$NEW_RESOURCE_VERSION" == "$SDK_RESOURCE_VERSION" ]; then
...
+		source "$script_path/update-version.sh"
+		exit 0;
+	fi
 else	 
 	# Update the SDK_VERSION
 	echo "Resource Version is not up to date. Changing major version."
-	NEW_MAJOR_VERSION="${NEW_RESOURCE_VERSION}001"
+	NEW_MAJOR_VERSION="v${NEW_RESOURCE_VERSION}001"
 	SDK_VERSION="v${NEW_MAJOR_VERSION}.0.0" 
+
```
matt-condon, line 35: Looks like you have `v` before both versions, I think one should be omitted

```suggestion
	NEW_MAJOR_VERSION="v${NEW_RESOURCE_VERSION}001"
	SDK_VERSION="${NEW_MAJOR_VERSION}.0.0" 

```

.github/workflows/autoupdate-prod.yaml
wtrocki: The key fix here was the bug we had for Major releases that prevented us from getting any breaking changes even in manual form. Since I made this improvement I decided to autogenerate them. 
If we see that autogeneration does have some issues we can always turn it off.

matt-condon: Is the manual major release approach still documented somewhere in case we do have some issues with autogeneration?

tools/releaser/scripts/new-version.sh
matt-condon: I see the comment is resolved but the repeated v is still present
Was I mistaken?
Looks like `SDK_VERSION` would evaluate to `vv2023-10-01001.0.0`

wtrocki: I could swear I have resolved that comment 1h ago. It might be that commit were lost or my resolution were ignored. 
Sorry for that.

matt-condon (approved):
LGTM

PR #217, opened 2023-10-04
fix: add separate commits for the automation workflow

wtrocki:
## Description

Adding separate commits for each phases of the automatic code generation

We committing changes in 3 phases:
1. Commit is an open API file and versions that we can review
2. The second one is generated code and docs.
3. For version bumps

## Verification

Workflow temporarily runs as part of this PR: https://github.com/mongodb/atlas-sdk-go/pull/218/files
Tested generation and releases end to end.

## Why

Breaking changes needs source code committed to calculate diff without raising issue


tools/releaser/scripts/breaking-changes.sh
```diff
@@ -3,15 +3,15 @@ set +e
...
 # Usually "${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha }}" 
 # TARGET_BREAKING_CHANGES_FILE - file to save breaking changes
 script_path=$(dirname "$0")
 
 echo "Installing go-apidiff"
-go install github.com/joelanford/go-apidiff@latest
+go install github.com/joelanford/go-apidiff@latest > /dev/null
 
-GIT_BASE_REF=${GIT_BASE_REF:-"origin/main"}
+GIT_BASE_REF=${GIT_BASE_REF:-git rev-parse head || echo}
```
wtrocki, line 14: This is much better for testing as we can then run workflow from separate branches with multiple corner cases without editing workflow

.github/workflows/autoupdate-prod.yaml
```diff
@@ -29,11 +35,14 @@ jobs:
...
+      - name: Commit Generator Changes
+        run: |
+          git add . && git commit -m "fix: Generated SDK source code and docs"
       - name: Release updates
         if: steps.verify-changed-files.outputs.files_changed == 'true'
         working-directory: ./tools
-        env: 
-          GIT_BASE_REF: origin/main
+        env:
+          GIT_BASE_REF: ${{ github.sha }}
```
wtrocki, line 45: I needed to make this change in order to test it from non main branch (like this PR)

matt-condon (approved):

PR #230, opened 2023-10-25
chore: Adjust node version for asdf, display versions

lantoli:
## Description

- Adjust node version for asdf
- Display npm and node version to understand what versions are being used as they're not explicitely set in gh actions.

## Type of change:

## Required Checklist:

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [X] I have added tests that prove my fix is effective or that my feature works

## Further comments

Example of output for npm version:
```
{
  'openapi-generation-tools': '0.1.0',
  npm: '9.8.1',
  node: '18.18.2',
  acorn: '8.10.0',
  ada: '2.6.0',
  ares: '1.19.1',
  brotli: '1.0.9',
  cldr: '43.1',
  icu: '73.2',
  llhttp: '6.0.11',
  modules: '108',
  napi: '9',
  nghttp2: '1.57.0',
  nghttp3: '0.7.0',
  ngtcp2: '0.8.1',
  openssl: '3.0.10+quic',
  simdutf: '3.2.14',
  tz: '2023c',
  undici: '5.26.3',
  unicode: '15.0',
  uv: '1.44.2',
  uvwasi: '0.0.18',
  v8: '10.2.154.26-node.26',
  zlib: '1.2.13.1-motley'
}
```

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,3 @@
 golang 1.20
-nodejs lts-hydrogen
+nodejs 18.18.2
+
--- a/tools/scripts/generate.sh
+++ b/tools/scripts/generate.sh
@@ -23,6 +23,7 @@ echo "# Running transformation based on $OPENAPI_FILE_NAME to the $transformed_f
 cp "$OPENAPI_FOLDER/$OPENAPI_FILE_NAME" "$openapiFileLocation"
 
 npm install
+npm version
 npm run sdk:transform -- "$openapiFileLocation"
 
 echo "# Running OpenAPI generator validation"
--- a/tools/scripts/generate.sh
+++ b/tools/scripts/generate.sh
@@ -23,7 +23,6 @@ echo "# Running transformation based on $OPENAPI_FILE_NAME to the $transformed_f
 cp "$OPENAPI_FOLDER/$OPENAPI_FILE_NAME" "$openapiFileLocation"
 
 npm install
-npm version
 npm run sdk:transform -- "$openapiFileLocation"
 
 echo "# Running OpenAPI generator validation"
--- a/tools/releaser/scripts/new-version.sh
+++ b/tools/releaser/scripts/new-version.sh
@@ -39,7 +39,6 @@ else
 
 	echo "Modifying all instances of version from $SDK_RESOURCE_VERSION to $NEW_RESOURCE_VERSION across the repository."
 	npm install
-	npm version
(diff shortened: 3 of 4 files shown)


tools/scripts/generate.sh
```diff
@@ -23,6 +23,7 @@ echo "# Running transformation based on $OPENAPI_FILE_NAME to the $transformed_f
 cp "$OPENAPI_FOLDER/$OPENAPI_FILE_NAME" "$openapiFileLocation"
 
 npm install
+npm version
```
lantoli, line 26: to understand what node version is being used when called from GH action without explicitly setting a version

.tool-versions
```diff
@@ -1,2 +1,3 @@
 golang 1.20
-nodejs lts-hydrogen
+nodejs 18.18.2
+
```
lantoli, line 3: it's a bad practice and supported in new asdf versions to refer to an alias in .tool-versions, you must use the full version: https://github.com/asdf-vm/asdf-nodejs
```
 In asdf, we prioritize reproducibility, so you cannot use partial versions or codenames in a .tool-versions file.
```

tools/scripts/generate.sh
wtrocki: Probably intention was to use:
```suggestion
npm --version
```

npm version is available from npm 6. We are running this job in evergreen with npm version not managed by us so we tend to avoid using latest features.

wtrocki: Context: only the minimal nodejs version set in the package.json matters. We do not care about the nodejs versions. Package versions are locked thru the package.lock file so adding this will be a noise and confuse others that npm/versions matter. In most of the cases, we going to print the same versions.

wtrocki: For clarity I think what we intend is:
```suggestion
`echo "DEBUG: script is running using $(npm --version)"
```

lantoli: thank you, i was not aware that it was called from other projects, will leave only the asdf change

wtrocki (approved):
LGTM awesome work!

tools/scripts/generate.sh
wtrocki: All good. We can still add some improvements to scripts but we need to:

- Make sure that our change is aligned with the existing lowest npm version we support
- Have solid "why" for the change to minimize blast radius

PR #241, opened 2023-11-20
fix: ignore tj-actions from frequent updates

wtrocki:
## Description

Opening for opinions


gssbzn (commented):
what about if instead of ignoring moving the frequency to monthly or quarterly?

wtrocki:
Good point. Reviewed other packages from that category. There are no drawback of keeping this in monthly schedule

PR #247, opened 2023-12-12
CLOUDP-217493: introduce pointer for arrays

wtrocki:
## Description

See CLOUDP-217493 for details.

## Requirements

- Case when empty array needs to be sent to the server
- Case when we do not want to send any data.
- Case when we send data in an array

## Testing

Run e2e examples manually.
```                                                                                                                                                                                                                                                          
▶ go run ./aws_cluster/aws.go                           
SDK: Created new MongoDB cluster: 65799f11d5ec4340bb0a1c5a 
Please wait up to 10 minutes for the cluster to provision.
```

## Documentation

Added documentation covering new behavior

## Risks

Changes would not trigger any **compilation issues** but will change behavior if the struct have an empty array defined.

Before:
```
struct property wasn't serialized - no update 
No ability to "reset" array to empty array.
```
After:
```
struct property is sent as empty array (will reset data)
developers need to remove field to restore old behaviour
```

Choosing this solution based on feedback from @lantoli. 

## Downstream impact

Changes need verification for impact on integrations and atlas cli 
For most of the projects, this change will require no-op on update.

To use this version run
```
go get go.mongodb.org/atlas-sdk/v20231115001@CLOUDP-217493-pointer
```


wtrocki:
Array of string doesn't behave the same way as array of structs for generator.

docs/doc_5_best_practices.md
```diff
@@ -28,6 +28,45 @@ if test != ""
     requestBody.StringPointerValue = test
 ```
 
+## Working with Arrays
```
lantoli, line 31: the examples are really with Slices, or we can say Working with Arrays and Slices

tools/config/go-templates/model_simple.mustache
```diff
@@ -228,7 +228,7 @@ func (o *{{{classname}}}) UnmarshalJSON(bytes []byte) (err error) {
 	{{#deprecated}}
 		// Deprecated
 	{{/deprecated}}
-		{{name}} {{^required}}{{^isArray}}{{^isFreeFormObject}}*{{/isFreeFormObject}}{{/isArray}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
+		{{name}} {{^required}}{{^isArray}}{{^isFreeFormObject}}*{{/isFreeFormObject}}{{/isArray}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}}{{^isArray}},omitempty{{/isArray}}{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
```
lantoli, line 231: nice how easy it is to change generated code!

wtrocki: That is only for this case. For array/slice pointers, we will require changes in several files and a lot of logic across different templates.

docs/doc_5_best_practices.md
```diff
@@ -28,6 +28,45 @@ if test != ""
...
+data := Data{
+    dataField: []string{}
+}
+```
+
+2. Update request without updating array field
+
+When performing update operation we recomend the struct property to not present. 
+
+> NOTE: Property definition as nil would trigger error or invalid behaviour
```
AgustinBettati, line 51: I am not sure I understand this statement. Does `Property definition as nil` refer to the following example?
```
data := Data{
    dataField: nil
}
```

If this is the case, what would be the risk? Does the marshalling include the property with a `null` value instead of a `[]`?

wtrocki: > If this is the case, what would be the risk?

Inconsistency in handling these values in Atlas for updates.

wtrocki: > Does the marshalling include the property with a null value instead of a []?

Yes.
Example:
https://go.dev/play/p/mRMMVjdNLwr

docs/doc_5_best_practices.md
```diff
@@ -28,6 +28,45 @@ if test != ""
     requestBody.StringPointerValue = test
 ```
 
+## Working with Arrays and Slices
+
+When working with arrays in Golang we need to consider various scenarios such as the absence of a struct property  or property being null.
```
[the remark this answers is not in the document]
wtrocki: Mostly because of semantics of go and Atlas API.

The problem can be seen in this playground example:
https://go.dev/play/p/a0y4DOHEHtX

That is the situation after this PR fix. 

https://go.dev/play/p/mRMMVjdNLwr

I have provided some context in PR description and thread offline.

wtrocki:
Closing and will work on alternative solution that will use pointers for arrays.

PR #253, opened 2024-01-04
CLOUDP-219156: Add ability for SDK to flag readonly and write only fields

lovisaberggren:
## Description

`ReadOnly` and `WriteOnly` fields are used in the Atlas OpenAPI file to flag what values should be sent or retrieved by the server. This PR adds this information to the SDK comments to flag these properties for fields that have them.

- [x] Template change
- [x] Regenerated code


wtrocki (approved):
LGTM

wtrocki:
To ensure that end users understand the meaning of this comment we can short documentation in the repository.

tools/config/go-templates/model_simple.mustache
```diff
@@ -17,6 +17,12 @@ type {{classname}} struct {
 {{#description}}
 	// {{{.}}}
 {{/description}}
+{{#isReadOnly}}
+  // Read only field
```
gssbzn, line 21: in go you should always end your docs with a full stop (.), also there's something odd with the spacing compared to the examples above

Also godoc thing, new lines not necessarily translate to new lines when rendered in https://pkg.go.dev/go.mongodb.org/atlas-sdk@v1.0.0/admin I would evaluate if you really want it as a new line or should be inline

lovisaberggren: Added the full stop and fixed the indentation.

Q on the other point: do you mean that it would result in something like:
```
// Some description. // Write only field.
```

gssbzn: > do you mean that it would result in something like

More like
> Some description. Write only field.

when rendered there's no comment markers

lovisaberggren: I think it's fine with it rendering on the same line 👍

gssbzn: works for me if it works for you

gssbzn (approved):

PR #257, opened 2024-01-10
CLOUDP-221984: remove redudant Mix max validation for SDK

wtrocki:
## Why

Client-side validation for min and max has been causing issues with compatibility and breaking changes in the past. 
We had similar issues with pattern matching and decided to remove regexp validation at early phases of the project.
Validation in the client duplicates backend validation and it is hard to handle on the client side.
This validation has been introduced by the upstream template and is generally safe to remove.
In SDK min and Max can be considered hints.

## What

Do no trigger errors as it would mean that changing min and max values causes breaking changes
This is one of the series of mitigations for min max and pattern values. 
We need to ensure that those values are correct and reflect reality so downstream fixes will be made as well.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-221984

## Type of change:

## Required Checklist:

### DIFF
--- a/admin/api_access_tracking.go
+++ b/admin/api_access_tracking.go
@@ -177,18 +177,6 @@ func (a *AccessTrackingApiService) listAccessLogsByClusterNameExecute(r ListAcce
 	localVarHeaderParams := make(map[string]string)
 	localVarQueryParams := url.Values{}
 	localVarFormParams := url.Values{}
-	if strlen(r.groupId) < 24 {
-		return localVarReturnValue, nil, reportError("groupId must have at least 24 elements")
-	}
-	if strlen(r.groupId) > 24 {
-		return localVarReturnValue, nil, reportError("groupId must have less than 24 elements")
-	}
-	if strlen(r.clusterName) < 1 {
-		return localVarReturnValue, nil, reportError("clusterName must have at least 1 elements")
-	}
-	if strlen(r.clusterName) > 64 {
-		return localVarReturnValue, nil, reportError("clusterName must have less than 64 elements")
-	}
 
 	if r.authResult != nil {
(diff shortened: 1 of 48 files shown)


oarbusi (approved):

admin/api_access_tracking.go
```diff
@@ -177,18 +177,6 @@ func (a *AccessTrackingApiService) listAccessLogsByClusterNameExecute(r ListAcce
 	localVarHeaderParams := make(map[string]string)
 	localVarQueryParams := url.Values{}
 	localVarFormParams := url.Values{}
-	if strlen(r.groupId) < 24 {
```
wtrocki, line 180: This type of validation was never part of any requirement or covered in spec. 
We inherited it through upstream. Errors that are reported aren't api errors - introducing unneeded complexity.

oarbusi (approved):

tools/config/go-templates/api.mustache
```diff
@@ -166,48 +166,6 @@ func (a *{{{classname}}}Service) {{#lambda.camelcase}}{{{nickname}}}{{/lambda.ca
 		return {{#returnType}}localVarReturnValue, {{/returnType}}nil, reportError("{{paramName}} is required and must be specified")
 	}
 	{{/isPathParam}}
-	{{#minItems}}
```
lantoli, line 169: would it be useful to have this info as doc or we'll be encouraging teams to remove this info from OpenAPI completely?

lantoli (approved):

tools/config/go-templates/api.mustache
wtrocki: No. min, max, and pattern are very valid information that can be used in the documentation and other form.
We need to make sure that this info is correct but also do not need to impose validation for it in SDK

matt-condon (approved):

PR #232, opened 2023-11-10
POC for mocking SDK

wtrocki:
## Description

Example how mocking can be automated in the SDK generation.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run `make fmt` and formatted my code

PR #262, opened 2024-01-19
CLOUDP-222817: make execute methods mockable

wtrocki:
## Description

We are adding ability to mock execute methods/interface in SDK.

Side effect: All interfaces methods must be public.
Since executor library uses interface we cannot have any private execute methods to be available on interface itself.

## Why

If we do not provide an public methods on interface generation of the mocks outside the SDK would not work as generators will not have access to private methods. 
Alternatively we could consider #232 as mocking solution in SDK but still some customers will likely prefer to use their own mocking.

## Review

Only *.mustache file matters. Rest of the changes are generations from mustache file

### DIFF
--- a/admin/api_access_tracking.go
+++ b/admin/api_access_tracking.go
@@ -35,7 +35,7 @@ type AccessTrackingApi interface {
 	ListAccessLogsByClusterNameWithParams(ctx context.Context, args *ListAccessLogsByClusterNameApiParams) ListAccessLogsByClusterNameApiRequest
 
 	// Interface only available internally
-	listAccessLogsByClusterNameExecute(r ListAccessLogsByClusterNameApiRequest) (*MongoDBAccessLogsList, *http.Response, error)
+	ListAccessLogsByClusterNameExecute(r ListAccessLogsByClusterNameApiRequest) (*MongoDBAccessLogsList, *http.Response, error)
 
 	/*
 		ListAccessLogsByHostname Return Database Access History for One Cluster using Its Hostname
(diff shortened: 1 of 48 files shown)


tools/config/go-templates/api.mustache
```diff
@@ -48,7 +48,7 @@ type {{classname}} interface {
 	{{nickname}}WithParams(ctx context.Context, args *{{operationId}}ApiParams) {{operationId}}ApiRequest
 
 	// Interface only available internally
```
gssbzn, line 50: is this still true?

wtrocki: Yes. Maybe I should change message to make method available for mocking

```suggestion
	// Method available only for mocking purposes
```

fmenezes: why do we need this method on the interface? can we keep it private and remove from interface? is it needed for mocking?

admin/api_access_tracking.go
```diff
@@ -132,7 +132,7 @@ func (r ListAccessLogsByClusterNameApiRequest) Start(start int64) ListAccessLogs
 }
 
 func (r ListAccessLogsByClusterNameApiRequest) Execute() (*MongoDBAccessLogsList, *http.Response, error) {
-	return r.ApiService.listAccessLogsByClusterNameExecute(r)
```
wtrocki, line 135: @fmenezes To be able to use mocks along with the implementation our execute method. using interface to call ListAccessLogsByClusterNameExecute() (which can be replaced by mock.

Any form of casting to implementation will also prevent mocking here. 
We inherited this API design from upstream.

tools/config/go-templates/api.mustache
wtrocki: Answered inline: https://github.com/mongodb/atlas-sdk-go/pull/262/files#r1459286558

gssbzn (approved):

PR #275, opened 2024-02-09
chore: Triggers Go proxy cache refresh

lantoli:
## Description

Triggers Go proxy cache refresh when a release is done

## Type of change:

## Required Checklist:

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added any necessary documentation (if appropriate)
- [X] I have run `make fmt` and formatted my code

### DIFF
--- a/.github/workflows/autorelease.yaml
+++ b/.github/workflows/autorelease.yaml
@@ -23,3 +23,9 @@ jobs:
           body: ${{ env.RELEASE_NOTES }}
           tag_name: ${{ env.RELEASE_TAG }}
           name: ${{ env.RELEASE_TAG }}
+      - name: Trigger Go proxy cache refresh
+        run: |
+          LATEST_SDK_TAG=$(curl -sSfL https://api.github.com/repos/mongodb/atlas-sdk-go/releases/latest | jq -r '.tag_name')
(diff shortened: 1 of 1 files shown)


.github/workflows/autorelease.yaml
```diff
@@ -23,3 +23,9 @@ jobs:
           body: ${{ env.RELEASE_NOTES }}
           tag_name: ${{ env.RELEASE_TAG }}
           name: ${{ env.RELEASE_TAG }}
+      - name: Trigger Go proxy cache refresh
+        run: |
+          LATEST_SDK_TAG=$(curl -sSfL https://api.github.com/repos/mongodb/atlas-sdk-go/releases/latest | jq -r '.tag_name')
```
lantoli, line 28: we could fo LATEST_SDK_TAG=${{ env.RELEASE_TAG }} but I prefer in this way so the step is more "independent" and we can check both versions are equal

wtrocki (approved):
LGTM

.github/workflows/autorelease.yaml
wtrocki: Wondering if `go get` is enough, but if you have tested this method then we good.

lantoli: this was recommended by @gssbzn

PR #287, opened 2024-03-23
feat: Adds test mock support with Mockery

lantoli:
## Description

We are adding Testify mocks generated by Mockery to make it easier to test use of SDK from your packages.

Example of use in Terraform Atlas Provider repo: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2069

## Review

Files in mockadmin folder are auto-generated.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run `make fmt` and formatted my code

PR #288, opened 2024-03-25
chore: Follow-up to Mockery followup

lantoli:
## Description

Follow-up for Mockery PR: https://github.com/mongodb/atlas-sdk-go/pull/287

- Uses Mockery binary (recommended) instead of go install
- Add mocking to contributing guide

## Type of change:

## Required Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,6 +84,7 @@ Repository contains following SDKs
 - `tools`: tooling used to generate Atlas Admin SDK. See [./tools](./tools) for more information.
 - `examples`: SDK examples
 - `internal`: folder for hosting non user facing SDK methods and core helpers
+- `mockadmin`: [Testify mocks](https://pkg.go.dev/github.com/stretchr/testify/mock) generated with [Mockery](https://github.com/vektra/mockery) for API interfaces
 
 ## VSCode debugging configuration
 VSCode developers might use following debugging configuration for transformer and go debugging
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -42,8 +42,10 @@ update-version:
 	## When manually updating major version
 	./releaser/scripts/update-version.sh
 
-.PHONY: generate_mocks
-generate_mocks:	
-	go install github.com/vektra/mockery/v2@v2.42.1
+.PHONY: generate_mocks_linux
+generate_mocks_linux:	
 	rm -Rf ../mockadmin
-	mockery
+	curl -L -o mockery.tar.gz https://github.com/vektra/mockery/releases/download/v2.42.1/mockery_2.42.1_Linux_x86_64.tar.gz
+	tar -xzf mockery.tar.gz
+	chmod +x mockery
+	./mockery
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -42,10 +42,6 @@ update-version:
 	## When manually updating major version
 	./releaser/scripts/update-version.sh
 
-.PHONY: generate_mocks_linux
-generate_mocks_linux:	
-	rm -Rf ../mockadmin
-	curl -L -o mockery.tar.gz https://github.com/vektra/mockery/releases/download/v2.42.1/mockery_2.42.1_Linux_x86_64.tar.gz
-	tar -xzf mockery.tar.gz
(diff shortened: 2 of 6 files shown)


tools/Makefile
```diff
@@ -42,8 +42,10 @@ update-version:
...
 	./releaser/scripts/update-version.sh
 
-.PHONY: generate_mocks
-generate_mocks:	
-	go install github.com/vektra/mockery/v2@v2.42.1
+.PHONY: generate_mocks_linux
+generate_mocks_linux:	
 	rm -Rf ../mockadmin
-	mockery
+	curl -L -o mockery.tar.gz https://github.com/vektra/mockery/releases/download/v2.42.1/mockery_2.42.1_Linux_x86_64.tar.gz
```
oarbusi, line 48: From [this comment](https://github.com/mongodb/atlas-sdk-go/pull/287#discussion_r1537256974) in the previous PR, this dependency would not be automatically updated, right?

lantoli: not sure I understand your question

oarbusi: Sorry, I meant that as far as I understand, @wtrocki wanted the mockery dependency version to be elsewhere so that it can be automatically updated by dependabot. As it is now, I don't think that's the case

wtrocki, line 48: There are couple antipatterns here:
1. We do not cache installation (mocking locally fails 
2. We do not clean up files
3. We do use inefficient and undocumented way of installation
4. We do not check if tool exist (making redundant installs without cache)

We tend to use `.bin` file in the repo so developers do not get tar.gz file as change when running this command.

See also ASDF docs: https://github.com/asdf-community/asdf-golang?tab=readme-ov-file#default-go-get-packages

Official docs:
https://vektra.github.io/mockery/latest/installation/

Seome examples of makefile efficient installations:

1. https://github.com/bxcodec/go-clean-arch/blob/e06c6d0cb37069b0ef56e3df67f80ca130a1ab82/misc/make/tools.Makefile#L85 
2. https://github.com/hyperledger/fabric/blob/111cff51600d26d4b4b05f52825da11e7629e971/gotools.mk#L36
3. https://github.com/tendermint/tendermint/blob/35581cf54ec436b8c37fabb43fdaa3f48339a170/scripts/mockery_generate.sh#L9-L15

CONTRIBUTING.md
```diff
@@ -84,6 +84,7 @@ Repository contains following SDKs
 - `tools`: tooling used to generate Atlas Admin SDK. See [./tools](./tools) for more information.
 - `examples`: SDK examples
 - `internal`: folder for hosting non user facing SDK methods and core helpers
+- `mockadmin`: [Testify mocks](https://pkg.go.dev/github.com/stretchr/testify/mock) generated with [Mockery](https://github.com/vektra/mockery) for API interfaces
```
wtrocki, line 87: IMHO we might need documentation that clarifies how we maintain mockery as part of the SDK.
Defining make target to regenerate locally and pointing to the tests

wtrocki (approved):
LGTM although I think makefile target should support both local and github runs.

tools/Makefile
wtrocki: Related to the https://github.com/mongodb/atlas-sdk-go/pull/288#discussion_r1537657532

Re version - I think we use pattern to define them in the header of the markdown to understand how to update them and cover them in the Contributing guide

lantoli: thank you, at the end I've followed the Docker approach so it works both in local and in the GH action

tools/Makefile
```diff
@@ -42,10 +42,6 @@ update-version:
...
-.PHONY: generate_mocks_linux
-generate_mocks_linux:	
-	rm -Rf ../mockadmin
-	curl -L -o mockery.tar.gz https://github.com/vektra/mockery/releases/download/v2.42.1/mockery_2.42.1_Linux_x86_64.tar.gz
-	tar -xzf mockery.tar.gz
-	chmod +x mockery
-	./mockery
+.PHONY: generate_mocks
+generate_mocks:	
+	./scripts/generate_mocks.sh
```
wtrocki, line 47: ❤️ Nice!

PR #328, opened 2024-05-17
task: update stable operations from the CLI

gssbzn:
## Description

We had not done this in a while, took some time to run it and update here


wtrocki (approved):
Code and docs needs regeneration

wtrocki:
PR needs to be regenerated. Adding Eng sync topic to discuss stable operations purpose as stability levels in SDK have been some source of confusion

PR #339, opened 2024-06-03
CLOUDP-252250:  extend validation to catch situations when overriden property has $ref

wtrocki:
## Description

CLOUDP-252250

Adding support for validating the cases where oneOf OpenAPI schema contains children with mixed $ref and type values. 

When both children have value with the same name we warn developers about potential duplicates.
If both types are different we fail validation. 
However validation did not check if mismatch can happen with $ref value.
This PR introduces fix to check $refs for type mismatch.

Examples:

```
21:53:19.646 WARN  Type missmatch found when merging base types. Ignoring due to known missmatches{
  "key": "eventTypeName",
  "firstValue": "object",
  "secondValue": "#/components/schemas/AlertAuditTypeView",
  "typeRefMismatch": true
}
21:53:19.644 WARN  Type missmatch found when merging base types. Ignoring due to known missmatches{
  "key": "units",
  "firstValue": "string",
  "secondValue": "#/components/schemas/TimeMetricUnits",
  "typeRefMismatch": true
}
21:53:19.644 WARN  Type missmatch found when merging base types. Ignoring due to known missmatches{
  "key": "latestDefinition",
  "firstValue": "#/components/schemas/SearchIndexDefinition",
  "secondValue": "#/components/schemas/TextSearchIndexDefinition",
  "typeRefMismatch": true
}
```

From the point of the upstream OpenAPI those values are correct however when we merge those in SDK we cannot represent them properly or perform any transformation.

## Type of change:

## Required Checklist:


tools/transformer/src/duplicate.ignore.json
```diff
@@ -0,0 +1,15 @@
+{
+        "ignoredProperties": [
+                "units",
```
wtrocki, line 3: Each will likely require follow up fix.

tools/transformer/src/transformations/oneOf.js
```diff
@@ -135,15 +136,30 @@ function handleDuplicates(parentObject, childObject) {
...
-        `${duplicatesSource} missmatch type detected: ${JSON.stringify(
-          missmatches,
-          undefined,
-          2
-        )}`
-      );
+      missmatches = missmatches.filter(missmatch =>{
+        for(const ignoredProperty of ignoredProperties){
+          if(missmatch.key == ignoredProperty ){
+            console.warn("Type missmatch found when merging base types. Ignoring due to known missmatches", JSON.stringify(
```
wtrocki, line 144: Once we triage all ignored properties I think we could move logging level to info to avoid excessive output in CI/CD.

tools/transformer/src/transformations/oneOf.js
```diff
@@ -135,15 +136,28 @@ function handleDuplicates(parentObject, childObject) {
   ]);
   if (duplicates.length > 0) {
     const duplicatesSource = childObject.title || "";
-    const missmatches = duplicates.filter((e) => e.typeMismatch);
+    let missmatches = duplicates.filter((e) => e.typeRefMismatch);
```
marcosuma, line 139: very nit (non-blocking): curious about the different spelling missmatch/mismatch - would be nice if we stayed consistent

marcosuma: is actually missmatch an allowed word? Maybe I am just ignoring it as a non-native speaker

wtrocki: Nice find. `Mismatch` is correct spelling.

wtrocki: Resolved

admin/model_cloud_provider_access_role.go
```diff
@@ -20,10 +20,10 @@ type CloudProviderAccessRole struct {
 	// Date and time when someone authorized this role for the specified cloud service provider. This parameter expresses its value in the ISO 8601 timestamp format in UTC.
 	// Read only field.
 	AuthorizedDate *time.Time `json:"authorizedDate,omitempty"`
-	// Date and time when someone created this role for the specified cloud service provider. This parameter expresses its value in the ISO 8601 timestamp format in UTC.  Alternatively: Date and time when this Azure Service Principal was created. This parameter expresses its value in the ISO 8601 timestamp format in UTC.
+	// Date and time when this Azure Service Principal was created. This parameter expresses its value in the ISO 8601 timestamp format in UTC.
 	// Read only field.
 	CreatedDate *time.Time `json:"createdDate,omitempty"`
-	// List that contains application features associated with this Amazon Web Services (AWS) Identity and Access Management (IAM) role.  Alternatively: List that contains application features associated with this Azure Service Principal.
+	// List that contains application features associated with this Azure Service Principal.
```
marcosuma, line 26: non blocking (just my understanding): why is the "alternatively" being removed? Was the previous code trying to merge two different documentations?

marcosuma (approved):
LGTM Thank you!

wtrocki:
Adding extra handling to improve state of the OpenAPI.

admin/model_cloud_provider_access_role.go
wtrocki: Great question. Initial design had merged description from all possible types that API can handle. In practice descriptions were confusing and very long. I have settled with base class description which usually reflects better what field mean. In some cases that base class would require small correction but it is still much better than having +60 lines description that does not mean much.

drinkbird (approved):

openapi/atlas-api-transformed.yaml
```diff
@@ -24146,74 +24064,8 @@ components:
...
-
-
-            Alternatively:
-
-            List of rules that determine whether MongoDB Cloud checks an object for the alert configuration. You can filter using the matchers array if the **eventTypeName** specifies an event for a host, replica set, or sharded cluster.
-
-
-            Alternatively:
-
-            List of rules that determine whether MongoDB Cloud checks an object for the alert configuration. You can filter using the matchers array if the **eventTypeName** specifies an event for a host, replica set, or sharded cluster.
```
wtrocki, line 24206: Good example of overly long schema and repeated descriptions.

wtrocki:
Verified:
- In backend CI/CD
- With Dev preview channel
- Production env

Overrides are possible in backend and Github repository at the same time

wtrocki:
Added documentation for the new validation internally including troubleshooting guide.

PR #342, opened 2024-06-06
CLOUDP-252756 - transform Search Schema

wtrocki:
## Description

CLOUDP-252756

Transformation restores previous state for the OpenAPI after recent major release.
Results will be visible in the generated OpenAPI release. I have tested transformation with different versions in OpenAPI.  I have designed it to automatically. 

Since this is one singular case we not going to make this an "official" transformation and keep it as placeholder to fix issue with search. We already have other placeholder in place at the moment.

### DIFF
--- a/tools/transformer/src/atlasTransformations.js
+++ b/tools/transformer/src/atlasTransformations.js
@@ -10,9 +10,7 @@ const {
   removeRefsFromParameters,
 } = require("./transformations");
 
-const {
-  resolveOpenAPIReference
-} = require("./engine/transformers");
+const { resolveOpenAPIReference } = require("./engine/transformers");
 
 const removeUnusedSchemas = require("./engine/removeUnused");
 
@@ -74,13 +72,18 @@ module.exports = function runTransformations(openapi) {
   return openapi;
 };
 
-// Workarounds for issues in the search API
+// Temporary transformation until new search version is introduced.
 function searchAPIIssuesTransformation(openapi) {
-  // API is overly complex in this case and provides no value as typed interface
   if (openapi.components.schemas.SearchIndexResponse) {
     const responseParent = openapi.components.schemas.SearchIndexResponse;
-    if (responseParent.properties && responseParent.properties.latestDefinition) {
-      if (responseParent.discriminator && responseParent.discriminator.mapping) {
+    if (
+      responseParent.properties &&
+      responseParent.properties.latestDefinition
+    ) {
+      if (
+        responseParent.discriminator &&
+        responseParent.discriminator.mapping
+      ) {
         responseParent.properties.latestDefinition = { oneOf: [] };
         for (const mappingKey in responseParent.discriminator.mapping) {
           const ref = responseParent.discriminator.mapping[mappingKey];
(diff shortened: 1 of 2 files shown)


tools/transformer/src/engine/transformers.js
```diff
@@ -127,10 +127,29 @@ function removeParentFromAllOf(child, parentName) {
   return true;
 }
 
+function resolveOpenAPIReference(openapi, ref) {
```
wtrocki, line 130: Common helper method for finding ref in openapi object

tools/transformer/src/atlasTransformations.js
```diff
@@ -67,3 +71,43 @@ module.exports = function runTransformations(openapi) {
...
 };
+
+// Workarounds for issues in the search API
+function searchAPIIssuesTransformation(openapi) {
+  // API is overly complex in this case and provides no value as typed interface
+  if (openapi.components.schemas.SearchIndexResponse) {
+    const responseParent = openapi.components.schemas.SearchIndexResponse;
+    if (
+      responseParent.properties &&
+      responseParent.properties.latestDefinition
```
gssbzn, line 82: do we have modern js available?
```suggestion
      responseParent?.properties?.latestDefinition
```

tools/transformer/src/atlasTransformations.js
```diff
@@ -67,3 +71,43 @@ module.exports = function runTransformations(openapi) {
...
+    ) {
+      if (
+        responseParent.discriminator &&
+        responseParent.discriminator.mapping
```
gssbzn, line 86: same as above

tools/transformer/src/atlasTransformations.js
```diff
@@ -67,3 +71,43 @@ module.exports = function runTransformations(openapi) {
...
+            continue; // Skip if there's no reference
+          }
+          const reference = resolveOpenAPIReference(openapi, ref);
+          responseParent.properties.latestDefinition.oneOf.push({
+            $ref:
+              reference.allOf &&
+              reference.allOf[1] &&
+              reference.allOf[1].properties &&
+              reference.allOf[1].properties.latestDefinition &&
+              reference.allOf[1].properties.latestDefinition.$ref,
```
gssbzn, line 101: same as above

gssbzn (approved):
LGTM to unblock happy to do as follow up

tools/transformer/src/atlasTransformations.js
wtrocki: Sadly we cannot make that assumption - due to internal system CI/CD. We had issues with `?` not being recognized.

tools/transformer/src/atlasTransformations.js
```diff
@@ -67,3 +71,42 @@ module.exports = function runTransformations(openapi) {
 
   return openapi;
 };
+
+// Temporary transformation until new search version is introduced.
```
marcosuma, line 75: nit: for the next time if there's a Jira issue (or perhaps even a link to another GH issue or PR) it could be great to link so that we remember the whole context

PR #348, opened 2024-06-19
CLOUDP-256771: add type mappings for supporting objects types

wtrocki:
## TL;DR 

This PR allows us to support **strong typing** of the otherwise untyped API and SDK methods

## Description

CLOUDP-256771 - Object types in the SDK should be threated as free form types. 

## Explanation

`map[string]interface{}` does not represent the OpenAPI `object` type properly. 
We cannot pass existing strongly typed structs to the SDK fields using `map[string]interface{}`  type.
`map[string]interface{}` type forces developers to transform their own structs leading to boilerplate mappers.
SDK users (e.g operator) would like to use their own types in this case - further simplifying end user experience.

Additionally 40% of the `map[string]interface{}` cases right now not only represent object but also base types and arrays.  

## Required context 

To reflect multiple values in the field OpenAPI provides `oneOf` operator. 
As architecture choice SDK merges all oneOf references into single struct
This produces limitation where oneOf with base type cannot be efficiently reflected in the SDK as 
Golang does not allow any type of polymorphism. For that case golang introduced `any` (`interface{}`) 
and offers out of the box support for parsing and casting of the different types.

## Impact

Move from `map[string]interface{}`to  `interface{}` allows users to supply:
- base types
- their own structs
- map[string]interface{} (backwards compatibility)

## Impact on the search API

Search API is widely untyped thus any existing untyped Search JSON files can be effectively read by SDK users (terraform, operator) and passed to SDK. IMHO this will improve general user experience as well.

## Backwards compatibility

Change would not cause any breaking changes.
See https://go.dev/play/p/qzi9VvtgaFH

## Documentation

Added to the PR. We will add additional info in release notes post merge.

## Alternatives

- Rewrite of SDK OpenAPI transformation engine to not merge types
- Create new virtual types and Implement custom golang parser handling specific types we want.

### DIFF
--- a/admin/model_atlas_search_analyzer.go
+++ b/admin/model_atlas_search_analyzer.go
@@ -141,7 +141,7 @@ func (o *AtlasSearchAnalyzer) GetTokenizer() map[string]interface{} {
 // and a boolean to check if the value has been set.
 func (o *AtlasSearchAnalyzer) GetTokenizerOk() (map[string]interface{}, bool) {
 	if o == nil {
-		return map[string]interface{}{}, false
+		return map[string]interface{}, false
 	}
 	return o.Tokenizer, true
 }
--- a/admin/model_atlas_search_analyzer.go
+++ b/admin/model_atlas_search_analyzer.go
@@ -9,20 +9,20 @@ import (
 // AtlasSearchAnalyzer struct for AtlasSearchAnalyzer
 type AtlasSearchAnalyzer struct {
 	// Filters that examine text one character at a time and perform filtering operations.
-	CharFilters *[]map[string]interface{} `json:"charFilters,omitempty"`
+	CharFilters *[]interface{} `json:"charFilters,omitempty"`
 	// Name that identifies the custom analyzer. Names must be unique within an index, and must not start with any of the following strings: - `lucene.` - `builtin.` - `mongodb.`
 	Name string `json:"name"`
 	// Filter that performs operations such as:  - Stemming, which reduces related words, such as \"talking\", \"talked\", and \"talks\" to their root word \"talk\".  - Redaction, which is the removal of sensitive information from public documents.
-	TokenFilters *[]map[string]interface{} `json:"tokenFilters,omitempty"`
+	TokenFilters *[]interface{} `json:"tokenFilters,omitempty"`
 	// Tokenizer that you want to use to create tokens. Tokens determine how Atlas Search splits up text into discrete chunks for indexing.
-	Tokenizer map[string]interface{} `json:"tokenizer"`
+	Tokenizer interface{} `json:"tokenizer"`
 }
 
 // NewAtlasSearchAnalyzer instantiates a new AtlasSearchAnalyzer object
 // This constructor will assign default values to properties that have it defined,
 // and makes sure properties required by API are set, but the set of arguments
 // will change when the set of required properties is changed
-func NewAtlasSearchAnalyzer(name string, tokenizer map[string]interface{}) *AtlasSearchAnalyzer {
+func NewAtlasSearchAnalyzer(name string, tokenizer interface{}) *AtlasSearchAnalyzer {
 	this := AtlasSearchAnalyzer{}
 	this.Name = name
 	this.Tokenizer = tokenizer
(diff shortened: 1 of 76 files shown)


lantoli:
not sure about this change, in general I think that the more specific the better, so map[string]interface{} is prefered to interface{} if we're representing a JSON object.    I'm not sure about changing this behavior because one specific struct

wtrocki:
Provided this PR as example of capabilities we have but no strong opinion on how to handle it. 

>  in general I think that the more specific the better, 

Does that mean that we want to introduce new custom openapi types for OpenAPI and SDK to handle?
`map[string]interface{}` cannot handle basic types like boolean. 

 Since this option provides:
 
 > their own structs
 
 We can effectively make our schema more strongly typed by documenting cases:
 
 ```
 // Use SomeObject model or boolean value
Field interface{} 
```

File: admin/model_search_index_update_request_definition.go [hunk 1]
```diff
@@ -16,7 +16,7 @@ type SearchIndexUpdateRequestDefinition struct {
 	// Method applied to identify words when searching this index.
 	SearchAnalyzer *string `json:"searchAnalyzer,omitempty"`
 	// Flag that indicates whether to store all fields (true) on Atlas Search. By default, Atlas doesn't store (false) the fields on Atlas Search.  Alternatively, you can specify an object that only contains the list of fields to store (include) or not store (exclude) on Atlas Search. To learn more, see Stored Source Fields.
-	StoredSource map[string]interface{} `json:"storedSource,omitempty"`
+	StoredSource interface{} `json:"storedSource,omitempty"`
```
oarbusi, line 19: Nice 👍

oarbusi (approved):
LGTM

admin/model_atlas_search_analyzer.go
```diff
@@ -141,7 +141,7 @@ func (o *AtlasSearchAnalyzer) GetTokenizer() map[string]interface{} {
 // and a boolean to check if the value has been set.
 func (o *AtlasSearchAnalyzer) GetTokenizerOk() (map[string]interface{}, bool) {
 	if o == nil {
-		return map[string]interface{}{}, false
+		return map[string]interface{}, false
```
marcosuma, line 144: @wtrocki just flagging for awareness

lantoli:
I would like to understand better the implications of this change.   I think it's better to have the more specific types the better, so fields that can be map[string]interface{} are preferred over interface{} that can represent anything.  
So in this case I'm not sure about changing all maps to any.

lantoli:
linter is failing

lantoli:
in the particular case of StoredSource, it sounds a bit strange to me to have a field that can be either true, false or an object.  I think this is the main issue.

for instance the team could change to have always an object and the field be part of it, e.g.:
```
"storedSource": { "storeAll": true  }  # equivalent to true
"storedSource": { "dontStoreOriginalDoc": true  }  # equivalent to false
```

more info: https://www.mongodb.com/docs/atlas/atlas-search/stored-source-definition/

```
true
Specifies that Atlas Search must store all the fields in the documents. Storing full documents might significantly impact performance during indexing and querying. To learn more, see Storing Source Fields.
false
Specifies that Atlas Search must not store the original source document. This is the default value for the storedSource option.
```

admin/model_atlas_search_analyzer.go
wtrocki: Ack! There is some extra work needed on the PR to make it final. Leaving it open for consideration and we can pick that up if needed.

wtrocki:
> in the particular case of StoredSource

OpenAPI does reflect all sorts of the formats that cannot be easily reflected in golang. 
We might think of some hacks.

gssbzn (approved):

wtrocki:
@lantoli Any impact on terraform/Integrations?

tools/scripts/generate.sh
```diff
@@ -33,6 +33,7 @@ npm exec openapi-generator-cli -- generate \
     -c "./config/config.yaml" -i "$openapiFileLocation" -o "$SDK_FOLDER" \
     --package-name="$client_package" \
     --type-mappings=integer=int \
+    --type-mappings=object=interface{} \
```
lantoli, line 36: can we better use `any` instead of `interface{}` which was added from Go 1.18 ?

lantoli (approved):
LGTM, just please use any instead of interface{}, thx for the PR!

lantoli:
after offline discussion I'm fine with leaving the PR as it is with interface{}

wtrocki:
@lantoli Than you for review.

PR #395, opened 2024-10-01
task: update min go version

gssbzn:
## Description

Update the min go version

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module go.mongodb.org/atlas-sdk/v20240805004
 
-go 1.21
+go 1.22.0
 
 require (
 	github.com/go-test/deep v1.1.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module go.mongodb.org/atlas-sdk/v20240805004
 
-go 1.21
+go 1.22.0
```
lantoli, line 3: i think it's better to just keep go 1.22. in go.mod, specific 1.22.0 , etc. would be in toolchain but I don't think that's a best practice

lantoli (approved):

go.mod
gssbzn: I haven't found clear guidance on full version here version language level and binary in the toolchain and in general what I've seen is no matter what you use the experience is really bad

wtrocki:
@lantoli We might still want to bump MMS version as currently job will download new golang every run instead of using evg provided ones.

PR #396, opened 2024-10-01
fix: fix warning on mixed (non)pointer receivers

gssbzn:
## Description

The template mixed pointer and non-pointer receivers and that causes some warnings

### DIFF
--- a/tools/config/go-templates/model_simple.mustache
+++ b/tools/config/go-templates/model_simple.mustache
@@ -172,14 +172,15 @@ func (o *{{classname}}) Set{{name}}(v {{vendorExtensions.x-go-base-type}}) {
 
 {{/required}}
 {{/vars}}
-func (o {{classname}}) MarshalJSONWithoutReadOnly() ([]byte, error) {
+func (o *{{classname}}) MarshalJSONWithoutReadOnly() ([]byte, error) {
 	toSerialize, err := o.ToMap()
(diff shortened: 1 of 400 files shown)


tools/config/go-templates/model_simple.mustache
```diff
@@ -172,14 +172,15 @@ func (o *{{classname}}) Set{{name}}(v {{vendorExtensions.x-go-base-type}}) {
 
 {{/required}}
 {{/vars}}
-func (o {{classname}}) MarshalJSONWithoutReadOnly() ([]byte, error) {
+func (o *{{classname}}) MarshalJSONWithoutReadOnly() ([]byte, error) {
```
gssbzn, line 175: this is the change, the rest is regenerating things

lantoli (approved):

tools/config/go-templates/model_simple.mustache
lantoli: i suppose we prefer this because performance?

gssbzn: you need pointers to modify the struct, which we do for setters, and you should not mix pointers and non pointers so stick to the one you need

Screenshot 2024-10-01 at 11 53 18

lantoli: thanks, i didn't think of the setters

PR #406, opened 2024-10-02
task: update example go

gssbzn:
## Description

Update the example, Go keeps trying to also update this file so I want to commit the change t avoid doing it by accident

### DIFF
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -1,6 +1,8 @@
 module go.mongodb.org/atlas-sdk/v20240805004/examples
 
-go 1.21
+go 1.22.0
+
+toolchain go1.23.1
 
 replace go.mongodb.org/atlas-sdk/v20240805004 => ../
 


wtrocki (approved):

examples/go.mod
```diff
@@ -1,6 +1,8 @@
 module go.mongodb.org/atlas-sdk/v20240805004/examples
 
-go 1.21
+go 1.22.0
+
+toolchain go1.23.1
```
lantoli, line 5: what are the pros/cons of defining this toolchain as opposed to not define it?

examples/go.mod
```diff
@@ -1,6 +1,8 @@
...
 
 replace go.mongodb.org/atlas-sdk/v20240805004 => ../
```
lantoli, line 7: @gssbzn what are the pros/cons of defining this toolchain as opposed to not define it?

gssbzn: toolchain is suppose to be for local dev vs as a lib

wtrocki: Yes. We always need to keep go version 2 versions behind latest but tooling can be latest (and we should align our internal system go version with tooling version)

lantoli: but this is for examples, can it be used as a library or it should be fine just to use go 1.23.1 in examples' go.mod?

wtrocki: Does not make difference for me as you properly pointed out sdk use case does not apply for examples.

PR #414, opened 2024-10-04
CLOUDP-276381: adding formatting to SDK

wtrocki:
## Description

Adding missing SDK error formatting introduced recently by us in backend.

## Review details

1. Review example file 
2. Tests should be passing (example is an test)
3. Review template

## Type of change:

## Required Checklist:

### DIFF
--- a/admin/client.go
+++ b/admin/client.go
@@ -531,7 +531,7 @@ func (c *APIClient) makeApiError(res *http.Response, httpMethod, httpPath string
 		newErr.error = err.Error()
 		return newErr
 	}
-	newErr.error = formatErrorMessage(res.Status, httpMethod, httpPath, v)
+	newErr.error = FormatErrorMessageWithDetails(res.Status, httpMethod, httpPath, v)
 	newErr.model = v
 	return newErr
 }
@@ -691,9 +691,15 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
 	e.error = errorString
 }
 
-// format error message using title and detail when model implements Error
-func formatErrorMessage(status, path, method string, v ApiError) string {
-	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v",
+// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
+func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
+	detailString := ""
+	if v.BadRequestDetail != nil {
+		badRequestDetail, _ := json.Marshal(v.GetBadRequestDetail())
+		detailString = string(badRequestDetail)
+	}
+
+	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v, BadRequestDetail: %v",
 		method, path, status, v.GetErrorCode(),
-		v.GetDetail(), v.GetReason(), v.GetParameters())
+		v.GetDetail(), v.GetReason(), v.GetParameters(), detailString)
 }
--- a/admin/client.go
+++ b/admin/client.go
@@ -691,7 +691,7 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
 	e.error = errorString
 }
 
-// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
+// FormatErrorMessageWithDetails formats error message using error struct fields. It should be only used for testing.
 func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
 	detailString := ""
 	if v.BadRequestDetail != nil {
--- a/admin/client.go
+++ b/admin/client.go
@@ -693,13 +693,13 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
 
 // FormatErrorMessageWithDetails formats error message using error struct fields. It should be only used for testing.
 func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
-	detailString := ""
+	badRequestDetailString := ""
 	if v.BadRequestDetail != nil {
 		badRequestDetail, _ := json.Marshal(v.GetBadRequestDetail())
-		detailString = string(badRequestDetail)
+		badRequestDetailString = string(badRequestDetail)
 	}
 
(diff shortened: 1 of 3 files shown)


admin/client.go
```diff
@@ -691,9 +691,15 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
...
 }
 
-// format error message using title and detail when model implements Error
-func formatErrorMessage(status, path, method string, v ApiError) string {
-	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v",
+// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
+func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
+	detailString := ""
+	if v.BadRequestDetail != nil {
+		badRequestDetail, _ := json.Marshal(v.GetBadRequestDetail())
```
lantoli, line 698: do we want to check for errors in case JSON is incorrect?

admin/client.go
```diff
@@ -691,9 +691,15 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
 	e.error = errorString
 }
 
-// format error message using title and detail when model implements Error
-func formatErrorMessage(status, path, method string, v ApiError) string {
-	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v",
+// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
```
lantoli, line 694:
```suggestion
// FormatErrorMessageWithDetails formats error message using error struct fields: Should be only used for testing
```

more Go idiomatic for func comments

admin/client.go
```diff
@@ -691,9 +691,15 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
...
+	detailString := ""
+	if v.BadRequestDetail != nil {
+		badRequestDetail, _ := json.Marshal(v.GetBadRequestDetail())
+		detailString = string(badRequestDetail)
+	}
+
+	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v, BadRequestDetail: %v",
 		method, path, status, v.GetErrorCode(),
-		v.GetDetail(), v.GetReason(), v.GetParameters())
+		v.GetDetail(), v.GetReason(), v.GetParameters(), detailString)
```
lantoli, line 704: knit: we have "details" name for two things, don't know if we could change one of them

tools/config/go-templates/client.mustache
```diff
@@ -652,9 +652,16 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
 	e.error = errorString;
 }
 
-// format error message using title and detail when model implements Error
-func formatErrorMessage(status, path, method string, v ApiError) string {
-	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v",
-			method, path, status, v.GetErrorCode(),
-	 		v.GetDetail(), v.GetReason(), v.GetParameters())
+// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
+func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
```
lantoli, line 656: i guess my suggestion for func comment should be changed here in the template

admin/client.go
```diff
@@ -691,9 +691,15 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
...
 
-// format error message using title and detail when model implements Error
-func formatErrorMessage(status, path, method string, v ApiError) string {
-	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v",
+// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
+func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
+	detailString := ""
+	if v.BadRequestDetail != nil {
+		badRequestDetail, _ := json.Marshal(v.GetBadRequestDetail())
+		detailString = string(badRequestDetail)
```
lantoli, line 699:
```suggestion
		if badRequestDetail, err := json.Marshal(v.GetBadRequestDetail()); err != nil {
		   detailString = string(badRequestDetail)
		}
```

shouldn't we check for errors converting JSON?

lantoli (commented):
LGTM, just a few minor comments, note that linter is failing

lantoli (approved):

admin/client.go
```diff
@@ -691,9 +691,15 @@ func (e *GenericOpenAPIError) SetError(errorString string) {
 	e.error = errorString
 }
 
-// format error message using title and detail when model implements Error
-func formatErrorMessage(status, path, method string, v ApiError) string {
-	return fmt.Sprintf("%v %v: HTTP %v (Error code: %q) Detail: %v Reason: %v. Params: %v",
+// FormatErrorMessageWithDetails: format error message using error struct fields: Should be only used for testing
+func FormatErrorMessageWithDetails(status, path, method string, v ApiError) string {
```
lantoli, line 695: can't it be used in prod code, e.g. to show Terraform errors to the user?

admin/client.go
wtrocki: We do parse JSON first from backend response. Even if we had invalid JSON  at this point somehow by regression in future would you see us doing different handling here? TL:DR - SDK does have request logger that shows rawa data. I do not like to throw errors from error handling method

admin/client.go
wtrocki: https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_2_error_handling.md 

With that context I do not think we need anyone to use that method directly. Exposing that method will give us ability to do better unit tests in cli/TF as we can autoformat error message (right now we have those values hardcoded which is maintenance burden every time test changes).

tools/config/go-templates/client.mustache
wtrocki: Yes. I always accept comment in 1 place and regenerate locally. 
Normal approach for SDK.

admin/client.go
wtrocki: Done

admin/client.go
lantoli: ok

lantoli (approved):

wtrocki:
Thank you for review @lantoli

PR #415, opened 2024-10-04
CLOUDP-277141: remove redundant methods that are not longer used

wtrocki:
## Description

CLOUDP-277141

Removing public methods:
- Methods were generated by upstream templates from OpenAPI generator. 
- Implementation was intended to be used internally by SDK (despite methods being public)
- Breaking changes will be generated automatically post merge of the method.
   - We going to issue major breaking change in the SDK

Link to any related issue(s): 

## Type of change:

## Required Checklist:


gssbzn (commented):
the methods are public how do we know they are unused?

lantoli (approved):
nice cleanup!

wtrocki:
> the methods are public how do we know they are unused?

Those methods do not have correct behaviour and are legacy from the community openapi generator. 

If we want to remove them we will need to issue breaking change/changelog. 
Posted PR for early review but need to put more work in docs to cover entire public API for models. 
Alternatively I considered deprecating those methods (and I'm looking for suggestions if we should/could deprecate other model methods)

lantoli:
@wtrocki @gssbzn  at least they're not used in MongoDB repos: https://github.com/search?q=org%3Amongodb+ToMap%28%29+language%3AGo&type=code&l=Go

but they could be used by some customers.  

As they shouldn't be used I think it's ok to delete them and create a new major version. (As a matter of fact I realized we were using it in Terraform but was not doing what we expected)

wtrocki:
I think we are guaranteed to get breaking change in the next release thus with additional documentation we should be on to remove those methods and issue breaking change notes.

PR #417, opened 2024-10-07
task: move unused codebase to the new package

wtrocki:
## Description

Cleanup in unused auth module. 
Auth Module were never used moving to subpackage. In the follow up PRs we will be introducing alternative authentication methods and we need auth module to be part of the subpackage.
For more info see: https://github.com/mongodb/atlas-sdk-go/pull/429

### DIFF
--- a/auth/code/device_flow.go
+++ b/auth/code/device_flow.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package auth
+package code
 
 import (
 	"context"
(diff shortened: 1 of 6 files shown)


lantoli (approved):
should it be deleted if not being used?

wtrocki:
@lantoli This happened because old manual SDK dependency is still used in the atlas CLI for Auth:
https://github.com/mongodb/mongodb-atlas-cli/blob/71ad153aad4ff3742c2e6f864045a443c3e275a8/internal/cli/auth/register.go#L26

Since more auth methods will be introduced soon, we need that code in subpackage.

PR #418, opened 2024-10-07
task: Updates link in Preview GHA

lantoli:
## Description

Updates link in Preview GHA to the latest version, so it will be automatically updated when new versions are created.

Link to any related issue(s):  CLOUDP-276361

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/autoupdate-preview.yaml
+++ b/.github/workflows/autoupdate-preview.yaml
@@ -56,7 +56,7 @@ jobs:
             ## Using early preview changes
 
             ```sh
-            go get go.mongodb.org/atlas-sdk/v20230201003@dev-latest   
+            go get go.mongodb.org/atlas-sdk/v20240805004@dev-latest   
             ```
 
             ## Automated checks done for preview sdk


.github/workflows/autoupdate-preview.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
             ## Using early preview changes
 
             ```sh
-            go get go.mongodb.org/atlas-sdk/v20230201003@dev-latest   
+            go get go.mongodb.org/atlas-sdk/v20240805004@dev-latest
```
AgustinBettati, line 59: I suspect we wont be keeping this up to date right? Maybe leaving a generic <major-version> can work as well

AgustinBettati (approved):

.github/workflows/autoupdate-preview.yaml
lantoli: files in this expression are automatically updated, including this file: https://github.com/mongodb/atlas-sdk-go/blob/CLOUDP-276361_preview_text/tools/releaser/scripts/version-paths.sh#L3

but at some moment this might be de-synch'ed,  from now it should be updated as with the other files

wtrocki: We fetching branch so any version will work but results depend on sdk version in integration.

In reality user will need to use sdk version from go.mod otherwise this command will add new sdk instead of replacing new one.

PR #423, opened 2024-10-08
chore: Includes build in each PR check

lantoli:
## Description

Includes build in each PR check.

Link to any related issue(s): CLOUDP-277647

## Type of change:

## Required Checklist:


.github/workflows/pr.yml
```diff
@@ -8,12 +8,11 @@ on:
       - main
   pull_request:
 jobs:
-  golangci-lint:
+  code-health:
```
lantoli, line 11: simplifying and joining linter and tests in the same job as their initialization is the same and it's what takes most of the time

.github/workflows/tools.yaml
lantoli, line 1: these checks are passed now in every PR

.github/workflows/pr.yml
wtrocki: I think someone had strong opinion about small jobs but forgot who had that :D
I'm ok with larger PR job.

lantoli: i agree with that general idea, but in this case they have exactly the same initialization and second one depends on first one, so I thought they could be together, most of the code here is for initiazalition (which is duplicated in each former job)

tools/config/go-templates/client.mustache
```diff
@@ -481,15 +481,15 @@ func (c *APIClient) makeApiError(res *http.Response, httpMethod, httpPath string
 
 	localVarBody, err := io.ReadAll(res.Body)
 	if err != nil {
-		newErr.error = err.Error()
+		newErr.error = fmt.Sprintf("(%s) failed to read response body: %s", res.Status, err.Error())
```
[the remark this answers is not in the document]
wtrocki: Thank you for doing it.

.github/workflows/pr.yml
lantoli: but happy to revert if you prefer that I leave them as they were

wtrocki (approved):
LGTM based on assumption that you have tested both tools linting (affected here) and regeneration. Job seems to be fast enough. In future PRs that affect CI/CD we can add extra notes to docs.

.github/workflows/pr.yml
wtrocki: I'm ok with the larger job

tools/config/go-templates/client.mustache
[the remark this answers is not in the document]
EspenAlbert: Thank you!
I didn't know about the templates, but I will remember next time 👍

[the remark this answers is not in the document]
lantoli: @EspenAlbert no need to remember, PR check will fail next time :-)

PR #425, opened 2024-10-08
chore: Delete unneeded templates

lantoli:
## Description

Delete unneeded templates as they don't contribute to the generated code.

Running this command from tools folder to make sure that file generation is kept unchanged: 
```
make clean_client generate_client generate_docs generate_mocks  generate_tests
```

Link to any related issue(s): CLOUDP-277683

## Type of change:

## Required Checklist:


tools/config/go-templates/model_simple.mustache
```diff
@@ -28,15 +28,8 @@ type {{classname}} struct {
 {{/deprecated}}
 	{{name}} {{^required}}{{^isFreeFormObject}}*{{/isFreeFormObject}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
 {{/vars}}
-{{#isAdditionalPropertiesTrue}}
```
lantoli, line 31: these isAdditionalPropertiesTrue blocks seem not to have any effect

auth/oauth.go
```diff
@@ -125,7 +125,7 @@ type TokenSource interface {
 	Token() (*Token, error)
 }
 
-func (c *Config) Do(ctx context.Context, req *http.Request, v interface{}) (*core.Response, error) {
+func (c *Config) Do(ctx context.Context, req *http.Request, v any) (*core.Response, error) {
```
lantoli, line 128: interface{} -> any leftover

docs/doc_5_best_practices.md
```diff
@@ -69,7 +69,7 @@ The golang `ìnterface{}` type allows you to use the following types as input ar
 
 - Basic types (integer, boolean)
 - Your own structures
-- map[string]interface{} for free-form objects
+- map[string]any for free-form objects
```
lantoli, line 72: interface{} -> any leftover

wtrocki: think free form are just any right?

lantoli: yes, although here I'm not changing doc, just interface{} to alias any

wtrocki:
@lantoli As discussed those templates are not used due to format of the OpenAPI - for example enum template. 
We had initial approach to have all OpenAPI generator templates in the repository.

OpenAPI generator have templates embedded. Templates we have in the repository override those templates ensuring that we do not have different versions of various template files present.

Some of the templates you have removed are used for testing (like `make generate_tests` that is very useful for us to verify correctness). Some are not executed because openapi does not have them. Despite removing them OpenAPI generator will still run them but this time completely different version that is not compatible with other templates. 
This has potential of regressions and further maintenance challenges. 

My personal recommendation is to split removal of the templates to separate PR and have required leftovers from previous PRs merged. WDYT?

tools/config/go-templates/api_test.mustache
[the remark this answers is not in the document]
wtrocki: `make generate_tests`

docs/doc_5_best_practices.md
wtrocki: `any for free-form objects` is the correct form. Do you mind applying it?

tools/config/go-templates/model_simple.mustache
wtrocki: Until they appear in the OpenAPI

wtrocki: What you see is conicidence that OpenAPI input does not trigger some of those claims. If that changes entire logic of SDK will be broken and we would need to investigate why. 
If your approach is to make SDK lean let's propose it separately as there would be some extra work required including validation for those changes to not break SDK in the future.

lantoli: there are 10 occurrences of "additionalProperties" in transformed OpenAPI file, e.g. [here](https://github.com/mongodb/atlas-sdk-go/blob/main/openapi/atlas-api-transformed.yaml#L22082)

So not sure if this logic is working fine or it's for other scenarios that are not happening yet. 

I don't think logic would break but that new scenario would be ignored and wouldn't be taken into account.

Anyway I'm thinking about pausing this PR and see first if it's feasible to upgrade generator version.

docs/doc_5_best_practices.md
lantoli: @wtrocki sorry I'm not sure I'm following you, can you give me an example where I should apply it?

lantoli:
@wtrocki yeah, i can split in different PRs.  Also I'm thinking about adding a note in the beginning of the files saying that they're not currently being used instead of deleting them.

Also  I'm thinking about pausing this PR and see first if it's feasible to upgrade generator version, then see if delete/comment unused template files.

tools/config/go-templates/model_simple.mustache
wtrocki: SGTM

docs/doc_5_best_practices.md
wtrocki:
```suggestion
- any for free-form objects
```

tools/config/go-templates/api_test.mustache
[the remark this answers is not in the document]
wtrocki: Running `make generate_tests` will give you example of the output.

lantoli:
at the end I've added a comment instead of deleting templates.

tools/config/go-templates/model_anyof.mustache
```diff
@@ -1,3 +1,5 @@
+// THIS TEMPLATE IS CURRENTLY NOT BEING USED
```
wtrocki, line 1: FYI: To be precise they are used by openapi generator.
Current model do not contain data that trigger output for template.

wtrocki (approved):

tools/config/go-templates/model_anyof.mustache
lantoli: thanks, adjusting message to:  THIS TEMPLATE IS NOT BEING TRIGGERED BY THE CURRENT MODEL

wtrocki: SGTM

PR #426, opened 2024-10-08
chore: Removes toMap and associated functions

lantoli:
## Description

Removes toMap and associated functions.

It was tested successfully with Terraform repo: https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/11241711780

Link to any related issue(s): CLOUDP-277767

## Type of change:

## Required Checklist:

### DIFF
--- a/admin/api_access_tracking.go
+++ b/admin/api_access_tracking.go
@@ -170,8 +170,8 @@ func (a *AccessTrackingApiService) ListAccessLogsByClusterNameExecute(r ListAcce
 	}
 
 	localVarPath := localBasePath + "/api/atlas/v2/groups/{groupId}/dbAccessHistory/clusters/{clusterName}"
-	localVarPath = strings.Replace(localVarPath, "{"+"groupId"+"}", url.PathEscape(parameterValueToString(r.groupId, "groupId")), -1)
-	localVarPath = strings.Replace(localVarPath, "{"+"clusterName"+"}", url.PathEscape(parameterValueToString(r.clusterName, "clusterName")), -1)
+	localVarPath = strings.Replace(localVarPath, "{"+"groupId"+"}", url.PathEscape(r.groupId), -1)
+	localVarPath = strings.Replace(localVarPath, "{"+"clusterName"+"}", url.PathEscape(r.clusterName), -1)
 
 	localVarHeaderParams := make(map[string]string)
 	localVarQueryParams := url.Values{}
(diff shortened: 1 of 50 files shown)


admin/api_access_tracking.go
```diff
@@ -170,8 +170,8 @@ func (a *AccessTrackingApiService) ListAccessLogsByClusterNameExecute(r ListAcce
 	}
 
 	localVarPath := localBasePath + "/api/atlas/v2/groups/{groupId}/dbAccessHistory/clusters/{clusterName}"
-	localVarPath = strings.Replace(localVarPath, "{"+"groupId"+"}", url.PathEscape(parameterValueToString(r.groupId, "groupId")), -1)
-	localVarPath = strings.Replace(localVarPath, "{"+"clusterName"+"}", url.PathEscape(parameterValueToString(r.clusterName, "clusterName")), -1)
+	localVarPath = strings.Replace(localVarPath, "{"+"groupId"+"}", url.PathEscape(r.groupId), -1)
```
lantoli, line 173: we're always passing a string not a struct, no need to use parameterValueToString

tools/config/go-templates/client.mustache
```diff
@@ -141,14 +126,6 @@ func parameterAddToHeaderOrQuery(headerOrQueryParams any, keyPrefix string, obj
 				value = "invalid"
 
 			case reflect.Struct:
-				if t,ok := obj.(modelWithMap); ok {
```
lantoli, line 144: toMap is not being implemented

wtrocki: None of the models implement it so this code does not execute. 
If we had any objects as parameters (currently that is not possible to express in swagger core) we would need to add JSON marshaler here. 

Separately. We initially had approach that minimized changes in the templates and flagged specific areas we change: https://github.com/mongodb/atlas-sdk-go/blob/main/tools/CONTRIBUTING.md#modification-of-the-custom-templates

This approach was to be able to continue update SDK with upstream. 
If your preference is to finetune templates we should likely remove {{! X-XGEN-MODIFIED }} and look to remove any dead code from templates (there is much more that we left in order to not make any unnecessary changes in templates.

tools/config/go-templates/api.mustache
```diff
@@ -154,7 +154,7 @@ func (a *{{{classname}}}Service) {{{nickname}}}Execute(r {{operationId}}ApiReque
 	}
 
 	localVarPath := localBasePath + "{{{path}}}"{{#pathParams}}
-	localVarPath = strings.Replace(localVarPath, "{"+"{{baseName}}"+"}", url.PathEscape(parameterValueToString(r.{{paramName}}, "{{paramName}}")), -1){{/pathParams}}
+	localVarPath = strings.Replace(localVarPath, "{"+"{{baseName}}"+"}", url.PathEscape(r.{{paramName}}), -1){{/pathParams}}
```
wtrocki, line 157: I assume that you have tested all the type parameters we have in the API

wtrocki (approved):
LGTM. Haven't done testing

tools/config/go-templates/api.mustache
lantoli: all TF tests passed and they have a good coverage.
also SDK wouldn't compile if some param were not a string, as it's required by strings.Replace

tools/config/go-templates/client.mustache
wtrocki: SGTM

tools/config/go-templates/api.mustache
wtrocki: That is odd. We do have integer and time params in API. It is just that in this structure they always represented as strings?  We definitely do have time based params. Very clever to use TF tests here

lantoli: note that this function is called only in path params, not in the body

wtrocki:
As discussed. I think we can merge this PR and have follow up jira to rethink the approach for templates. 
Changes will improve code to not check for Object parameters as our API does not have them.

tools/config/go-templates/api.mustache
wtrocki: Yes. Upstream SDK supports **Object** based parameters but they are rare in openapi and do not exist in Atlas.

PR #429, opened 2024-10-09
CLOUDP-278691: Credentials authentication

wtrocki:
## Summary

PR adds new The OAuth client manages tokens by retrieving and refreshing them as needed using the _client_credentials_ flow. It checks token validity based on expiration and refreshes expired tokens automatically by interacting with the OAuth server. The client ensures a valid token is always available before making API requests, handling errors like expired or invalid tokens gracefully. The implementation provides a reliable and automated way to maintain authentication.

## Jira

CLOUDP-278691

## Notes

Implementation will support all use cases for community, customers and our own internal clients.
Documentation will be provided in separate PR to get docs team approval.
PR only provides new package for alpha testing. Admin SDK integration will be done in follow up PRs.

## Testing

1. Unit tests
2. Example application.
```
export MONGODB_ATLAS_CLIENT_ID="your_client_id"
export MONGODB_ATLAS_CLIENT_SECRET="your_client_secret"
export MONGODB_ATLAS_URL=https://cloud.mongodb.com
go run ./examples/auth/credentials.go
```

### Diagram

```mermaid
graph TD
    A[Start] --> B[Token Request]
    B --> C{Is Token Expired?}
    C -- No --> D[Use Existing Token]
    C -- Yes --> E[Request New Token]
    E --> F[Atlas Admin API /token Request]
    F --> G[Return New Token]
    G --> H[Save Token in TokenSource]
    H --> I[Use New Token]
    D --> J[Send Authenticated API Request]
    I --> J
    J --> K[End]
```
 
### OAuth Client Implementation:

1. **Token Handling:**
   - The client relies on OAuth tokens to authenticate API requests. The tokens are issued via an Atlas API and are managed by the client.
   - A `TokenSource` is used to retrieve, store, and refresh tokens as needed. The `TokenSource` provides a JWT that includes an expiration time (`exp`).

3. **Token Expiration and Refresh Logic:**
   - When a token is fetched, the client checks its expiration date.
   - If the token is valid and not expired, it uses the existing token.
   - If the token is expired, the client automatically requests a new token from the OAuth server using the `client_credentials` grant type.

4. **Integration with OAuth Server:**
   - The client interacts with a mock OAuth server for token requests and refreshes. The mock server responds with a new token when requested via POST with a valid `grant_type`.
   - The client sends the necessary credentials (`ClientID`, `ClientSecret`) when making these requests.

5. **Token Validity:**
   - The client validates the token by checking its structure and ensuring it includes a valid `exp` claim.
   - If the token is near expiration, the client preemptively refreshes it to avoid any request failures.

6. **Error Handling:**
   - The client handles various scenarios such as expired tokens and invalid token responses.
   - It gracefully handles invalid tokens by refreshing them, and ensures a valid token is always available before sending API requests.


admin/atlas_client.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/mongodb-forks/digest"
+	"go.mongodb.org/atlas-sdk/v20240805004/auth/credentials"
```
wtrocki, line 11: reusing existing package with mutiple separate packages to use

auth/credentials/service_account.go
```diff
@@ -0,0 +1,157 @@
...
+	// Proceed with the underlying transport
+	return t.underlyingTransport.RoundTrip(req)
+}
+
+// SetAuthHeader sets the Authorization header with the access token.
+func (t *Token) SetAuthHeader(r *http.Request) {
+	r.Header.Set("Authorization", "Bearer " + t.AccessToken)
+}
+
+const expiryDelta = 10 * time.Second
```
wtrocki, line 116: We do not want to issue request for token that is about to expire in 10 seconds.

examples/auth/credentials.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// Fetch clientID and clientSecret from environment variables
+	clientID := os.Getenv("MONGODB_ATLAS_CLIENT_ID")
+	clientSecret := os.Getenv("MONGODB_ATLAS_CLIENT_SECRET")
+
+	if clientID == "" || clientSecret == "" {
+		log.Fatal("Missing CLIENT_ID or CLIENT_SECRET environment variables")
+	}
+	url := os.Getenv("MONGODB_ATLAS_URL")
```
lovisaberggren, line 33:
```suggestion
```
Seems to be already saved to var `host`

lovisaberggren (approved):
LGTM overall, nice work!

lovisaberggren (approved):
TY!

auth/credentials/transport.go
```diff
@@ -0,0 +1,27 @@
...
+// OAuthCustomHTTPTransport is a custom HTTP transport that injects the OAuth2 token into requests.
+type OAuthCustomHTTPTransport struct {
+	underlyingTransport http.RoundTripper
+	client              *OAuthClient
+}
+
+// RoundTrip implements the RoundTripper interface.
+func (t *OAuthCustomHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+    // Get a valid token (refreshing it if necessary)
+    token, err := t.client.getValidToken() // Get or refresh the token
```
fmenezes, line 17: nit: indentation seems off

fmenezes:
Your implementation checks out I wonder should you add context to the functions? Especially on the calls to save/load tokens, operations to disk might take longer than memory.

auth/credentials/transport.go
wtrocki: Awesome find! Checking with golang ci lint is not failing for missing gofmt

wtrocki:
>  I wonder should you add context to the functions? 

Great comment. Need to get more context for the change.

1. You mean to the Storage save/load interface? 
2. To clarify: So the idea is to reuse File the same context as HTTP calls use?
2.1 I initially assumed that e.g `AtlasCLIFileStorage` can have context on their own on structure,
and we do not need to share the same with OAuth client or expose it on Storage interface.

auth/credentials/api.go
```diff
@@ -0,0 +1,62 @@
...
+	// Custom context
+	Context *context.Context
+	// Custom base url for change
+	BaseURL *string
+}
+
+
+// NewServiceAccountOAuthClient initializes an OAuthClient with client credentials and custom TokenSource
+// Use this method to initialze custom token storage (filesystem)
+func NewServiceAccountOAuthClientWithTokenSource(opts ServiceAccountOAuthClientWithTokenSource) *OAuthClient {
```
lantoli, line 23: knit: it's idiomatic to start the comment with the func name

auth/credentials/api.go
```diff
@@ -0,0 +1,62 @@
...
+	//nolint:gosec //url only
+	tokenURL := "https://cloud.mongodb.com/api/oauth/token"
```
lantoli, line 25: can we avoid the linter exception by having a constant for default base url?

auth/credentials/transport.go
wtrocki: It looks like gofmt does not support wildcards in the format as other tools:

gofmt: The Go formatter tool.
-s: Simplifies the code by removing unnecessary whitespace and comments.
-w: Writes the changes back to the original files.
./**/**/*.go: A glob pattern that matches all .go files in the current directory and its subdirectories, recursively.


```
Projects/cloud/atlas-sdk-go  sa-poc ✔                                                                                                                                        5h5m  
▶ make fmt
==> Fixing source code with gofmt...
gofmt -s -w  ./**/*.go
// zero changes

▶ gofmt -s -w  ./**/**/*.go 
// All code formatted
``` 
🙈

fmenezes:
for context google's oauth mechanism decided to not have a cache load/store implementation (similar to your token source) and decided to leave it to custom transports (https://github.com/golang/oauth2/issues/8 && )

The only difference from your design would be instead of passing TokenSource to the client it would be passing something similar to the Transport, this way you might not have to provide more than InMemory and let the integration figure out what is best. It seems inline with other protocols.

auth/credentials/api.go
```diff
@@ -0,0 +1,62 @@
+package credentials
```
lantoli, line 1: nit: credentials and auth seems similar terms, what about using serviceaccount or similar  as package name? in that way func could be serviceaccount.NewClientWithTokenSource, NewHTTPClient, etc.

auth/credentials/oauth_client.go
```diff
@@ -0,0 +1,159 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// Token represents the internal OAuth2 token structure
```
lantoli, line 15:
```suggestion
// token represents the internal OAuth2 token structure
```

auth/credentials/oauth_client_test.go
```diff
@@ -0,0 +1,182 @@
...
+		// Create mock token response
+		tokenResponse := TokenResponse{
+			AccessToken:  token,
+			TokenType:    "Bearer",
+			ExpiresIn:    3600,
+		}
+
+		// Encode and return the JSON response
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(tokenResponse)
```
lantoli, line 41: maybe do a require.NotError or something similar to check err?

auth/credentials/transport.go
wtrocki: We also have disabled formatter on golang-ci-lint for formatter. Will get that enabled in separate PR.
@fmenezes thank you for spotting it. Reformatted source code and fixed make target.

auth/credentials/api.go
wtrocki: gosec linter works on strings.
Separately: Do you prefer to have constant? Leave 👍 and would do it 💪

auth/credentials/oauth_client_test.go
wtrocki: 💪 Can do! This code is mocked part of the test. We can print error etc.

auth/credentials/api.go
lantoli: if linter exception can't be removed then it's ok as it is, I would maybe change it a little bit but just a matter of preference, you can leave it as it is

```
baseURL := "https://cloud.mongodb.com"
if opts.BaseURL != nil {
		baseUrl = *opts.BaseURL 
	}
tokenURL := baseURL +  "/api/oauth/token"
```

auth/credentials/api.go
wtrocki: Pinged you offline on the documentation to discuss potential renames. We can apply any name we would decide upon post merge.

wtrocki:
> The only difference from your design would be instead of passing TokenSource to the client 

Yes. I have inspired on the OAuth2 library NewClient
https://pkg.go.dev/golang.org/x/oauth2#NewClient

The difference is that T

fmenezes:
> 1. You mean to the Storage save/load interface?
Yes, ideally disk operations should use context

> 2. To clarify: So the idea is to reuse the same context in FileStorage as HTTP calls use?
>    2.1 I initially assumed that e.g `AtlasCLIFileStorage` can have context on their own on structure,
>    and we do not need to share the same with OAuth client or expose it on Storage interface:
Context is meant to be shared to all operations to lifetime of an operation, if your operation does multiple things, say call an API, read something from memory and save to disk, ideally the same context is passed everywhere

Context as a param from struct is antipattern [ref](https://pkg.go.dev/context#:~:text=Do%20not%20store%20Contexts%20inside,func%20DoSomething%28ctx%20context.) (believe me I understand your pain)

fmenezes:
> Yes. I have inspired on the OAuth2 library NewClient
> https://pkg.go.dev/golang.org/x/oauth2#NewClient

The difference here is your implementation of Source does 2 things, provides the value and caches the updated value, on their case caching (updating it to disk or memory) is done at transport side, my point was it makes sense to move the caching part (SaveToken) to be a parameter to transport

lantoli (approved):

auth/credentials/api.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// NewServiceAccountOAuthClientWithTokenSource initializes an OAuthClient with client credentials and custom TokenSource
+// Use this method to initialze custom token storage (filesystem)
+func NewServiceAccountOAuthClientWithTokenSource(opts ServiceAccountOAuthClientWithTokenSource) *OAuthClient {
+	var tokenURL string
+	if opts.BaseURL != nil {
+		tokenURL = *opts.BaseURL + tokenAPIPath
+	} else {
+		tokenURL = serverURL + tokenAPIPath
```
marcosuma, line 35: not sure if I am misreading this, but isn't serverURL already = "https://cloud.mongodb.com" + tokenAPIPath ?

wtrocki:
Working on addressing @fmenezes comments. Moving PR to WIP.

auth/credentials/api.go
wtrocki: Code have been updated. Thank you

wtrocki:
## Update

Changed interface based on @fmenezes feedback:
- Renamed OAuthClient to TokenSource to replicate oauth2 like interface
- Simplified api.go to reflect golang standards
- Added package information clarifying that module is only for internal use (at the time of development until we have work finished)
- Enabled Users to create their own Transports (transport based use cases native to golang) for advanced integrations use cases (atlas CLI)

Change intends to support advanced use cases we see in Atlas CLI (https://github.com/mongodb/mongodb-atlas-cli/pull/3339). **No impact on Terraform/CDK and other clients that currently using SDK provided HttpClients and helpers.**

fmenezes:
Changes LGTM

lovisaberggren (approved):

gssbzn (commented):
why are we re implementing https://github.com/golang/oauth2?

PR #431, opened 2024-10-14
CLOUDP-272624: fixed example

wtrocki:
## Description

Added example for testing.

### DIFF
--- a/examples/mock/cluster_test.go
+++ b/examples/mock/cluster_test.go
@@ -1,4 +1,4 @@
-package test
+package examples
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


examples/mock/cluster_test.go
```diff
@@ -1,4 +1,4 @@
-package test
+package examples
```
wtrocki, line 1: FYI @lantoli This was developed as test but were put into examples package. 
See: https://go.dev/doc/tutorial/add-a-test

wtrocki:
Thank you for the review.

PR #457, opened 2024-10-24
CLOUDP-279931: Add ability to control user agent for token endpoints

lovisaberggren:
## Description

Added support for default and custom User Agent in newly introduced `OAuthTokenSource`. Also moved now shared default User Agent, DefaultCloudURL and ClientName to the `core` package for sharing between the Admin client and the token endpoints.

## Jira

[CLOUDP-279931](https://jira.mongodb.org/browse/CLOUDP-279931)

## Testing

Extended `credentials/oauth_test.go` for testing default and custom User Agent.

### DIFF
--- a/auth/credentials/api.go
+++ b/auth/credentials/api.go
@@ -2,6 +2,7 @@ package credentials
 
 import (
 	"context"
+	"go.mongodb.org/atlas-sdk/v20240805005/internal/core"
 	"net/http"
 )
 
@@ -11,13 +12,15 @@ import (
 const tokenAPIPath = "/api/oauth/token"
 
 // serverURL for atlas API
-const serverURL = "https://cloud.mongodb.com" + tokenAPIPath
+const serverURL = core.DefaultCloudURL + tokenAPIPath
 
 // AtlasTokenSourceOptions provides set of input arguments
 // for creation of credentials.TokenSource interface
 type AtlasTokenSourceOptions struct {
 	ClientID     string
 	ClientSecret string
+	// Custom user agent. core.DefaultUserAgent being default
+	UserAgent string
 	// Custom Token source. InMemoryTokenCache being default
 	TokenCache LocalTokenCache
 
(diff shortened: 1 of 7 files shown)


auth/credentials/api.go
```diff
@@ -28,14 +31,20 @@ type AtlasTokenSourceOptions struct {
 }
 
 // NewTokenSourceWithOptions initializes an OAuthTokenSource with advanced credentials.AtlasTokenSourceOptions
-// Use this method to initialize custom OAuth Token Cache (filesystem).
+// Use this method to initialize custom OAuth Token Cache (filesystem) or custom User Agent header.
```
wtrocki, line 34: As this method evolves that comment might no longer be relevant. I made it with notion to flag potential use case but we will handle it by hand written docs.
```suggestion
```

auth/credentials/oauth_test.go
```diff
@@ -186,3 +187,48 @@ func TestCustomTransport_RoundTrip(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, http.StatusOK, resp.StatusCode)
 }
+
+// Test default User Agent
+func TestOAuthClient_DefaultUserAgent(t *testing.T) {
+	// Assert custom User Agent
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.NotNil(t, r.Header.Get("User-Agent"))
```
wtrocki, line 195: ❤️ Awesome to see you made non-trivial unit tests case to test customization.

internal/core/client.go
```diff
@@ -22,6 +22,16 @@ import (
...
+)
+
+const (
+	// DefaultCloudURL is default base URL for the services.
+	DefaultCloudURL = "https://cloud.mongodb.com"
+	// ClientName of the v2 API client.
+	ClientName = "go-atlas-sdk-admin"
+	// DefaultUserAgent is default user agent header.
+	DefaultUserAgent = ClientName + "/" + Version + " (" + runtime.GOOS + ";" + runtime.GOARCH + ")"
 )
```
wtrocki, line 35: [super nit] client.go purpose was to separate HTTP client implementation for auth module. 
For those we can have constants.go just in case we would like to remove client.go along with the legacy code that uses it.

wtrocki (approved):
Super LGTM! Thank you for the fix.

auth/credentials/api.go
lovisaberggren: Yeah makes sense, I can remove it 👍

PR #459, opened 2024-10-24
CLOUDP-280007: Auth Revoke + Error cases + testing

wtrocki:
## Description

Revocation API provides way to "logout out" OAuth Token.

## Changes

Added unit tests covering non-happy path scenarios. 
Moved networking logic to a separate file for clarity.
Added revoke API with tests and examples.

## Verification

- Unit tests
- Run example:
```
▶ go run ./examples/auth/credentials.go    
{"links":[{"href":"https://cloud-dev.mongodb.com/api/atlas/v2/groups?includeCount=true&pageNum=1&itemsPerPage=1","rel":"self"}],"results":[{"clusterCount":0,"created":"2024-10-14T09:16:41Z","id":"670ce177f33cea212ea202a5","links":[{"href":"https://cloud-dev.mongodb.com/api/atlas/v2/groups/670ce177f33cea212ea202a5","rel":"self"}],"name":"test","orgId":"6707a96bc2b066602a4dfdcc","tags":[]}],"totalCount":1}
```
- Atlas CLI integration 
```
▶  ./bin/atlas logout --profile auth-test
Logged out with API key
```


auth/credentials/api.go
```diff
@@ -35,9 +38,11 @@ type AtlasTokenSourceOptions struct {
 // NewTokenSourceWithOptions initializes an OAuthTokenSource with advanced credentials.AtlasTokenSourceOptions
 func NewTokenSourceWithOptions(opts AtlasTokenSourceOptions) TokenSource {
 	var tokenURL string
+	var revokeUrl string
 	if opts.BaseURL != nil {
 		baseUrlNoSuffix := strings.TrimSuffix(*opts.BaseURL, "/")
 		tokenURL = baseUrlNoSuffix + tokenAPIPath
+		revokeUrl = baseUrlNoSuffix + revokeAPIPath
```
lantoli, line 45: empty revokeUrl if opts.BaseURL == nil ?

wtrocki: Adding test case.

wtrocki: Added case and verified it.

auth/credentials/oauth.go
```diff
@@ -71,50 +101,7 @@ func (c *OAuthTokenSource) refreshToken() (*Token, error) {
...
-	token := &Token{
-		AccessToken: tokenResp.AccessToken,
-		ExpiresIn:   tokenResp.ExpiresIn,
-		Expiry:      time.Now().Add(time.Duration(tokenResp.ExpiresIn) * time.Second),
-	}
-	return token, nil
-}
-
-// Additional time for Access Tokens to not expire.
+// ExpiryDelta Additional time for Access Tokens to not expire.
```
lantoli, line 104:
```suggestion
// ExpiryDelta is the Additional time for Access Tokens to not expire.
```

auth/credentials/oauth.go
```diff
@@ -151,7 +138,7 @@ func parseToken(accessToken string) (*Token, error) {
 
 	expiry := time.Unix(tokenData.Exp, 0)
 	if time.Now().After(expiry) {
-		return nil, errors.New("Token has expired")
+		return nil, errors.New("atlas cloud Access Token has expired")
```
lantoli, line 141:
```suggestion
		return nil, errors.New("Atlas Cloud Access Token has expired")
```

auth/credentials/oauth.go
```diff
@@ -160,3 +147,96 @@ func parseToken(accessToken string) (*Token, error) {
 		ExpiresIn:   int(time.Until(expiry).Seconds()),
 	}, nil
 }
+
+// fetchToken makes a manual POST request to Server (tokenUrl) to fetch the access Token.
```
lantoli, line 151:
```suggestion
// fetchTokenFromRemoteServer makes a manual POST request to Server (tokenUrl) to fetch the access Token.
```

auth/credentials/oauth.go
```diff
@@ -31,11 +39,33 @@ type OAuthTokenSource struct {
...
+	if err != nil {
+		return err
+	}
+	if tokenString != nil && *tokenString != "" {
+		err := c.revokeTokenInRemoteServer(*tokenString)
+		if err != nil {
+			return err
+		}
+
+		// Remove token from cache
```
lantoli, line 59: nit: i normally prefer code comments about WHY instead of WHAT we're doing. for example in this case we might want to have a tokenCache.RemoveToken or something similar:

```

func (s *MyTokenCache) RemoveToken(ctx context.Context) error {
   return s.SaveToken(ctx, "")
}
```

auth/credentials/oauth.go
```diff
@@ -160,3 +147,96 @@ func parseToken(accessToken string) (*Token, error) {
...
+	}
+	req.SetBasicAuth(c.clientID, c.clientSecret)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("User-Agent", c.userAgent)
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
```
lantoli, line 224: we may want to have an utility func for this is we're doing it multiple times

auth/credentials/oauth_test.go
```diff
@@ -100,7 +101,7 @@ func TestNewServiceAccountOAuthClientWithTokenSource_WithValidToken(t *testing.T
 
 // Test for providing valid Token and parsing the expiration date
 func TestNewServiceAccountOAuthClientWithTokenSource_WithWithExpiredToken(t *testing.T) {
-	// Generate a mock JWT Token with future expiration
+	// Generate a mock JWT Token with past expiration
```
lantoli, line 104: similar to previous comment about code comment, can we change code so comment is not needed? e.g. using var name `pastExpiration` or `expirationInPast`

auth/credentials/oauth_test.go
```diff
@@ -123,6 +124,31 @@ func TestNewServiceAccountOAuthClientWithTokenSource_WithWithExpiredToken(t *tes
 	assert.Equal(t, remoteToken, oAuthTokenSource.token.AccessToken)
 }
 
+// Test for providing valid Token and parsing the expiration date
+func TestNewServiceAccountOAuthClientWithTokenSource_WithWithAlmostExpiredToken(t *testing.T) {
+	// Generate a mock JWT Token that is going to expire in 5 seconds
+	expiration := time.Now().Add(-5 * time.Second)
```
lantoli, line 130: similar to previous comment, e.g. expirationAboutToHappen.


also these tests seems very similar, maybe we can have only one test func with multiple test cases like : 
``` 
	testCases := map[string]struct {
		expiration time.Time
		...
	}
             "almost expirer": ...,
	}

```

lantoli (dismissed):

auth/credentials/oauth.go
wtrocki: > WHY instead of WHAT 
```suggestion
		// Revoked token can be removed from cache
```

>  for example in this case we might want to have a tokenCache.RemoveToken

I think this is a separate suggestion to extend the interface. 
I'm going to move from `*string` to `string` which will make interface more clear and less prone to confusion. 
Remove will be redundant in this case.

auth/credentials/oauth.go
wtrocki: FYI: this would normally fail linting but we do not have error/comments linting enabled in the repository.

auth/credentials/oauth_test.go
wtrocki: SGTM. I have created jira to reduce duplication in tests.

auth/credentials/oauth.go
```diff
@@ -31,11 +39,33 @@ type OAuthTokenSource struct {
...
 	tokenCache   LocalTokenCache
 	ctx          context.Context
 }
 
+func (c *OAuthTokenSource) RevokeToken() error {
+	tokenString, err := c.tokenCache.RetrieveToken(c.ctx)
+	if err != nil {
+		return err
+	}
+	if tokenString != nil && *tokenString != "" {
```
lantoli, line 53: i guess string instead of *string would also help here

lantoli (approved):

auth/credentials/oauth.go
wtrocki: Exactly. I will do dedicated PR for this change to make sure that we do not have any leftovers in the docs etc.

PR #461, opened 2024-10-28
CLOUDP-279930: Admin SDK helper with examples

wtrocki:
## Description

CLOUDP-27993 - Adding Admin API support and separate example for supported use cases.
Documentation will be supplied separately.


wtrocki:
Moved to draft as changes depend on #459 to be merged first. Once that PR is merged we will move PR out of draft.
Early feedback appreciated.

examples/auth_advanced/advanced_client.go
```diff
@@ -89,7 +95,7 @@ type MyTokenCache struct {
 	mu          sync.Mutex
 }
 
-func (s *MyTokenCache) RetrieveToken(ctx context.Context) (*string, error) {
+func (s *MyTokenCache) RetrieveToken(_ context.Context) (*string, error) {
```
lantoli, line 98: why return *string instead of string?

wtrocki: Good comment. I will move interface from `*string` to `string`. This will mean we will use `""` as unset token instead of nil.

wtrocki: Changed

lovisaberggren (approved):

PR #471, opened 2024-11-04
CLOUDP-282641: Use oauth2 token source 

wtrocki:
## Description

Change to implementation to reuse oauth2 token source internally to perform token requests.
For discussion/spec see: CLOUDP-282641

## Progress

- [x] Updates to the codebase
- [x] Unit tests
- [x] Examples 
- [x] Documentation 
- [x] Custom User Agent


auth/credentials/api.go
```diff
@@ -63,40 +65,58 @@ func NewTokenSourceWithOptions(opts AtlasTokenSourceOptions) TokenSource {
 		ctx = *opts.Context
 	}
 
-	if opts.TokenCache == nil {
-		opts.TokenCache = &InMemoryTokenCache{}
+	config := &clientcredentials.Config{
```
wtrocki, line 68: Using oauth2 config to create oauth2.TokenSource

auth/credentials/api.go
```diff
@@ -63,40 +65,58 @@ func NewTokenSourceWithOptions(opts AtlasTokenSourceOptions) TokenSource {
...
 	return &http.Client{
 		Transport: &OAuthCustomHTTPTransport{
 			UnderlyingTransport: http.DefaultTransport,
-			TokenSource:         client,
+			TokenSource:         tokenSource,
 		},
 	}
 }
+
+// RevokeToken checks if TokenSource is a RevocableTokenSource and revokes Token.
```
wtrocki, line 115: Since `oauth2.TokenSource` does not have revocation, we're casting `RevocableTokenSource`

wtrocki: Actually. I will push a commit that will use RevocableTokenSource as internal TokenSource. We can always revert it.

auth/credentials/oauth.go
```diff
@@ -66,142 +64,82 @@ func (c *OAuthTokenSource) RevokeToken() error {
...
-
-	return &Token{
+	return &oauth2.Token{
 		AccessToken: accessToken,
 		Expiry:      expiry,
-		ExpiresIn:   int(time.Until(expiry).Seconds()),
 	}, nil
 }
 
-// fetchTokenFromRemoteServer makes a manual POST request to Server (tokenUrl) to fetch the access Token.
```
wtrocki, line 151: This is currently done by oauth2 lib. 
With this change we're missing the ability to set User-Agent and control error handling, retry logic etc.

auth/credentials/oauth.go
```diff
@@ -11,41 +11,39 @@ import (
...
-
-// Token represents the internal OAuth2 Token structure
-type Token struct {
-	AccessToken string    `json:"access_token"`
-	Expiry      time.Time `json:"expiry,omitempty"`
-	ExpiresIn   int       `json:"expires_in"`
-}
 
-// TokenSource interface allows to fetch and revoke valid OAuth Access Tokens.
-type TokenSource interface {
```
gssbzn, line 24: for simplicity you can do 
```go 
type TokenSource = oauth2.TokenSource
```

wtrocki: SGTM. TokenSource is not exposed externally (only used internally). 
This line removes it. What will be the purpose of redefining token source in this case?

auth/credentials/api.go
```diff
@@ -63,40 +65,52 @@ func NewTokenSourceWithOptions(opts AtlasTokenSourceOptions) TokenSource {
...
-		opts.TokenCache = &InMemoryTokenCache{}
+	// Override oauth2 http client
+	ctx = context.WithValue(ctx, oauth2.HTTPClient, newUserAgentHTTPClient(userAgent))
+
+	config := &clientcredentials.Config{
+		ClientID:     opts.ClientID,
+		ClientSecret: opts.ClientSecret,
+		TokenURL:     tokenURL,
+		AuthStyle:    oauth2.AuthStyleInHeader,
+		Scopes:       []string{}, // No support for scopes.
```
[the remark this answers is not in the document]
wtrocki: I prefer to flag `// No support for scopes.` in the wrapper to avoid potential help ticket asking to expose that in admin sdk. Happy to remove it.

auth/credentials/oauth.go
gssbzn: it is, it's a public package with an exported name, if you really don't want people to use these types or methods then make them all unexported or `/internal` and we can move on

auth/credentials/oauth_test.go
```diff
@@ -167,12 +174,10 @@ func TestOAuthClient_FetchToken_Failure(t *testing.T) {
...
 
-	// Call GetAccessToken expecting an error
-	_, err := tokenSource.GetValidToken()
+	// Call GetValidToken expecting an error
+	_, err := tokenSource.Token()
 	assert.Error(t, err)
-	oAuthTokenSource := tokenSource.(*OAuthTokenSource)
-	assert.Equal(t, mockServer.URL+tokenAPIPath, oAuthTokenSource.tokenURL)
-	assert.Contains(t, err.Error(), "Failed to obtain Access Token when fetching new OAuth Token from remote server")
+	assert.Contains(t, err.Error(), "oauth2: cannot fetch token: 401 Unauthorized")
```
lantoli, line 180: nit: you can use assert.ErrorContains

wtrocki: Nice! Yes. Good protip.

auth/credentials/oauth.go
wtrocki: We are exporting `RevocableTokenSource` in api.go. oauth2.Token source is only used as interface we inherit from.
That line removes TokenSource. Good point on not exposing only needed API. 
I will make oAuthTokenSource private as it is not needed to be exposed directly.

auth/credentials/oauth_test.go
```diff
@@ -218,10 +223,10 @@ func TestOAuthClient_FetchToken_FailureHtmlContentType(t *testing.T) {
...
-	// Call GetAccessToken expecting an error
-	_, err := client.GetValidToken()
+	// Call GetValidToken expecting an error
+	_, err := client.Token()
 	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "Failed to obtain Access Token when fetching new OAuth Token from remote server")
+	assert.Contains(t, err.Error(), "oauth2: cannot fetch token: 403 Forbidden")
 }
 
 // Test CustomTransport to ensure Token is injected into requests
```
lantoli, line 232: not changed in this PR, but prefer t.Cleanup to defer in tests, e.g.: https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests

go.sum
```diff
@@ -10,6 +10,8 @@ github.com/stretchr/objx v0.5.2 h1:xuMeJ0Sdp5ZMRXx/aWO6RZxdr3beISkG5/G/aIRr3pY=
 github.com/stretchr/objx v0.5.2/go.mod h1:FRsXN1f5AsAjCGJKqEizvkpNtU+EGNCLh3NxZ/8L+MA=
 github.com/stretchr/testify v1.9.0 h1:HtqpIVDClZ4nwg75+f6Lvsy/wHu+3BoSGCbBAcpTsTg=
 github.com/stretchr/testify v1.9.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
+golang.org/x/oauth2 v0.23.0 h1:PbgcYx2W7i4LvjJWEbf0ngHV6qJYr86PkAV3bXdLEbs=
+golang.org/x/oauth2 v0.23.0/go.mod h1:XYTD2NtWslqkgxebSiOHnXEap4TF09sJSc7H1sXbhtI=
```
lantoli, line 14: should we say something in the doc that we depend on this module?

lantoli (dismissed):

auth/credentials/api.go
```diff
@@ -22,81 +19,22 @@ const serverTokenURL = core.DefaultCloudURL + tokenAPIPath
...
-// Use this method for performing requests using http.DefaultTransport.
-// For more advanced use cases please create your own credentials.OAuthCustomHTTPTransport.
-func NewHTTPClientWithOAuthToken(client TokenSource) *http.Client {
-	return &http.Client{
-		Transport: &OAuthCustomHTTPTransport{
-			UnderlyingTransport: http.DefaultTransport,
-			TokenSource:         client,
-		},
-	}
+//func NewTokenSource(clientID, clientSecret string) RevocableTokenSource {
```
gssbzn, line 24: not sure if this still needed

admin/atlas_client.go
```diff
@@ -58,24 +60,11 @@ func UseDigestAuth(apiKey, apiSecret string) ClientModifier {
 // credentials.LocalTokenCache can be supplied to reuse OAuth Token across application restarts.
 // Warning: for advanced use cases please use credentials.NewTokenSource directly in your code pass it to UseHTTPClient method.
 // Warning: any previously set httpClient will be overwritten. To fully customize HttpClient use UseHTTPClient method.
-func UseOAuthAuth(clientID, clientSecret string, tokenCache credentials.LocalTokenCache) ClientModifier {
+func UseOAuthAuth(clientID, clientSecret string, tokenCache *oauth2.Token) ClientModifier {
```
gssbzn, line 63: this should probably accept a context and not sure if other modifiers like the ones for the base url

gssbzn:
if you don't need to load a cache token
```go
package main

import (
	"context"
	"fmt"
	"time"

	"log"

	atlas "go.mongodb.org/atlas-sdk/v20241023001/admin"
	"go.mongodb.org/atlas-sdk/v20241023001/auth/credentials"
)

func main() {
	conf := credentials.Config{
		RevokeURL: "http://localhost:8080/api/oauth/revoke",
	}
	conf.ClientID = "mdb_sa_id_<>"
	conf.ClientSecret = "mdb_sa_sk_<>"
	conf.TokenURL = "http://localhost:8080/api/oauth/token"
	ctx := context.Background()
	// t, err := conf.TokenSource(ctx).Token()
	// if err != nil {
	// 	log.Fatalln(err.Error())
	// }

	tc := conf.Client(ctx)

	client, err := atlas.NewClient(atlas.UseHTTPClient(tc), atlas.UseBaseURL("http://localhost:8080/"))
	if err != nil {
		log.Fatalln(err.Error())
	}
	projects, _, err := client.ProjectsApi.ListProjects(ctx).Execute()
	if err != nil {
		log.Fatalln(err.Error())
	}

	for _, p := range projects.GetResults() {
		time.Sleep(5 * time.Second)
		fmt.Println("projectID", p.GetId())
		clusters, _, err := client.ClustersApi.ListClusters(ctx, p.GetId()).Execute()
		if err != nil {
			log.Fatalln(err.Error())
		}
		for _, c := range clusters.GetResults() {
			fmt.Println("clusterName", c.GetName())
		}
	}
	err = conf.RevokeToken(ctx, t)
	if err != nil {
		log.Fatalln(err.Error())
	}
}

```

admin/atlas_client.go
wtrocki: This concept was inherited from old SDK, and it does hold us back due to `UseBaseURL` or `UseHttpClient` overriding effects of this one.

admin/atlas_client.go
```diff
@@ -58,24 +60,11 @@ func UseDigestAuth(apiKey, apiSecret string) ClientModifier {
...
-		} else {
-			tokenSource = credentials.NewTokenSourceWithOptions(credentials.AtlasTokenSourceOptions{
-				ClientID:     clientID,
-				ClientSecret: clientSecret,
-				BaseURL:      &c.Servers[0].URL,
-			})
-		}
-		httpClient := credentials.NewHTTPClientWithOAuthToken(tokenSource)
+		oauth := credentials.NewConfig(clientID, clientSecret)
+		source := oauth2.ReuseTokenSource(tokenCache, oauth.TokenSource(context.Background()))
```
wtrocki, line 66: [Q] Why using `ReuseTokenSource` from oauth2.

auth/credentials/clientcredentials.go
```diff
@@ -0,0 +1,78 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// Config describes a 2-legged OAuth2 flow, with both the
+// client application information and the server's endpoint URLs.
+type Config struct {
+	clientcredentials.Config
```
wtrocki, line 18: Do you have strong opinion on using config?
I'm concerned that this produces end user api that will expose too much.
Some examples:
```
	        // Token URL should be inherited from SDK. For all users will be hardcoded and not exposed to users.
		TokenURL:     "https://cloud.mongodb.com/oauth/token",
		// Scopes are present but they are not supported by our backend
		Scopes:       []string{"your-scope"},
            // By default will make redundant wrong requests against our API.
	      AuthStyle: oauth2.Auto
```

auth/credentials/clientcredentials.go
```diff
@@ -0,0 +1,78 @@
...
+	RevokeURL string
+	userAgent string
+}
+
+// tokenSource manages the OAuth Token fetching and refreshing using a LocalTokenCache.
```
wtrocki, line 23: `LocalTokenCache` seems to have been removed?

auth/credentials/api.go
```diff
@@ -22,81 +19,22 @@ const serverTokenURL = core.DefaultCloudURL + tokenAPIPath
...
+//		ClientSecret: clientSecret,
+//		UserAgent:    core.DefaultUserAgent,
+//	})
+//}
+
+func NewConfig(clientID, clientSecret string) *Config {
+	c := &Config{}
+	c.ClientID = clientID
+	c.ClientSecret = clientSecret
+	c.RevokeURL = revokeAPIPath
```
wtrocki, line 36: How this is supported by Token library?

wtrocki, line 36: Should I use path subtraction or replace this to be whole url?

auth/credentials/api.go
```diff
@@ -22,81 +19,22 @@ const serverTokenURL = core.DefaultCloudURL + tokenAPIPath
...
-	}
+//func NewTokenSource(clientID, clientSecret string) RevocableTokenSource {
+//	return NewTokenSourceWithOptions(AtlasTokenSourceOptions{
+//		ClientID:     clientID,
+//		ClientSecret: clientSecret,
+//		UserAgent:    core.DefaultUserAgent,
+//	})
+//}
+
+func NewConfig(clientID, clientSecret string) *Config {
```
wtrocki, line 32: Any reason we need to return oauth2.Config as end user API. Could we create token source here?

auth/credentials/clientcredentials.go
```diff
@@ -0,0 +1,78 @@
...
+			"Failed to revoke Access Token when fetching new OAuth Token from remote server",
+			resp.Header.Get("www-authenticate"))
+		return errors.New(formattedMessage)
+	}
+	return nil
+
+}
+
+func (c *tokenSource) Token() (*oauth2.Token, error) {
+	return c.conf.Token(c.ctx)
```
wtrocki, line 77: [Question] MIssing info how user agent is passed to token source.

admin/atlas_client.go
gssbzn: I'm not talking about the method but that this method should probably accept a ctx and should maybe allow to override the base url, of the oauth client which right now it doesn't allow me and why I could not really test this

admin/atlas_client.go
gssbzn: because it has the logic to check if the token is valid and reuse it or request a new one and keep using it while valid, I don't think we need to implement that logic ourselves

auth/credentials/api.go
gssbzn: we are returning a wrapper for `oauth2.Config` not `oauth2.Config` directly, the reason is that revoking is not part of TokenSource is part of Config and if you return TokenSource you loose access to revoke

auth/credentials/api.go
gssbzn: yeah bug, fixing

auth/credentials/clientcredentials.go
gssbzn: yes it has the implementation to get tokens, if we don't use it you have to implement them which was the whole point of this change, not re-implement oauth

to your concerns there's NewConfig which encapsulates this logic to create the one you want, so following your comments of "users won't use this package directly, then they won't use this struct directly right?

on top of that for this one 
>// By default will make redundant wrong requests against our API.

on my tests I haven't' seen this to be true

auth/credentials/clientcredentials.go
gssbzn: correct

admin/atlas_client.go
wtrocki: UseBaseURL overrides URL global. That is why `&c.Servers[0].URL` were used in the code that was removed.
Context should be also reused across the entire client otherwise we will have very confusing end user API.

wtrocki: Adding those changes.

auth/credentials/api.go
wtrocki: Not needed. Current approach using headless TokenSource (We create new TokenSource every time token is requested and use PersistentStore to reuse TokenSources.

admin/atlas_client.go
wtrocki: Prior to your commit, wrapper of ReuseToken source provided the ability to save OAuth token.

auth/credentials/clientcredentials.go
```diff
@@ -28,7 +29,7 @@ type tokenSource struct {
 
 // RevokeToken revokes the Access Token while also removing it from the Access Token Cache.
 // When the Access Token is expired or missing, revoke will return without any action.
-func (c *Config) RevokeToken(ctx context.Context, t *oauth2.Token) error {
+func (c *Config) RevokeToken(ctx context.Context, t *auth.Token) error {
```
wtrocki, line 32: We removed logic that accepted token string. 
To pass token we will require client to store expiration or parse token into auth.Token.

auth/credentials/clientcredentials.go
wtrocki: > if we don't use it you have to implement them which was the whole point of this change, not re-implement oauth

Prior to last commit, all changes we had reused `oauth.TokenSource.Token()` but also incorporated additional requirements from the scope: 
- Ability to integrate with Atlas CLI by creating transport instead of Client.
- Token being string. 
- Custom user agent etc. 
- RevokableTokenSource that provided immutable access to Token and Revoke so it can be part of the context without security risks.   

I like how lean and clean current approach is but I might need to reintroduce some removed code in order to support those features. Is that ok?

> on my tests I haven't' seen this to be true

We have provided override when creating config, but we are still exposing that field to end users which is my point.

wtrocki:
@gssbzn Thank you for changes. I will take from there. Might need to reintroduce some of the code to fill the gaps and request your review.

admin/atlas_client.go
gssbzn: >UseBaseURL overrides URL global. That is why &c.Servers[0].URL were used in the code that was removed.

Does that mean I need to know I have to call that function first?

auth/credentials/clientcredentials.go
gssbzn: yes, I'm not a fan of the previous approach parsing the string to guess the expiry date, specially when that parsing wasn't doing any sort of signature validation of the token so they could easily be tampered strings, I strongly belive that we shouldn't be dealing with strings but a Token object, this is similar to what the official oauth lib does

admin/atlas_client.go
wtrocki: Yes. That is why I think we should:
- Deprecate UseDigest 
- Change examples to UseHttpClient instead (for both digest and OAuth)

WDYT?

admin/atlas_client.go
gssbzn: yes for file systems, for in memory there's nothing to save, it just works

auth/credentials/clientcredentials.go
wtrocki: That means that customer applications will need to be dealing with building Token (from storage etc.)
Tampered strings/Token do not matter as all oauth backend does provide security features in this case. 
Client side uses expiration to avoid sending expired tokens only.

auth/credentials/clientcredentials.go
gssbzn: I don't see any of the points you mention captured in the scope

auth/credentials/clientcredentials.go
gssbzn: the token type is (de)serializable, it'll be safer to operate within the struct than string parsing of the token and risk a tampered one, even if the back-end would still catch it

admin/atlas_client.go
gssbzn: no strong opinion, the CLI already works with `UseHttpClient`

auth/credentials/clientcredentials.go
wtrocki: SGTM

admin/atlas_client.go
wtrocki: I think with this PR we moving from:

- SDK provides simple and comprehensive 
- Just use oauth2 + our solution for token storage we provided.

gssbzn: yes, which should've always been the case :)

auth/credentials/clientcredentials.go
wtrocki: Decision. We will bring previous change to supply user agent properly.

auth/credentials/clientcredentials.go
wtrocki: Covered offline. We will settle with oauth2.Token as representation of token. Apply custom user agent.
We do not need RevocableTokenSource as plan is to contribute those changes to auth2 in long run.

auth/credentials/transport.go
```diff
@@ -31,6 +32,27 @@ func (t *OAuthCustomHTTPTransport) RoundTrip(req *http.Request) (*http.Response,
...
 }
+
+// userAgentTransport that adds the User-Agent header
+type userAgentTransport struct {
+	underlyingTransport http.RoundTripper
+	userAgent           string
+}
+
+func (t *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("User-Agent", t.userAgent) // Set the custom User-Agent
```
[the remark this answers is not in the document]
wtrocki: Done

go.sum
wtrocki: We do not want to expose that information. We wrap all oauth2 fields.

auth/credentials/api.go
wtrocki: Removed

admin/atlas_client.go
wtrocki: Added method for saving token in filesystem

wtrocki:
## Update

1. Included all changes by @gssbzn 
2. Added way to send user agent when requesting token and store token when it is refreshed
3. Updated examples and unit tests.

admin/atlas_client.go
```diff
@@ -55,27 +57,35 @@ func UseDigestAuth(apiKey, apiSecret string) ClientModifier {
...
 	return func(c *Configuration) error {
-		var tokenSource credentials.TokenSource
-		if tokenCache != nil {
-			tokenSource = credentials.NewTokenSourceWithOptions(credentials.AtlasTokenSourceOptions{
-				ClientID:     clientID,
-				ClientSecret: clientSecret,
-				TokenCache:   tokenCache,
-				BaseURL:      &c.Servers[0].URL,
+		oauth := credentials.NewConfig(clientID, clientSecret)
+		ctx := context.WithValue(context.Background(), oauth2.HTTPClient,
```
gssbzn, line 64: should we be passing the context as and argument?

admin/atlas_client.go
```diff
@@ -55,27 +57,35 @@ func UseDigestAuth(apiKey, apiSecret string) ClientModifier {
...
+		source := oauth.TokenSource(ctx)
+		httpClient := oauth2.NewClient(ctx, source)
+		c.HTTPClient = httpClient
+		return nil
+	}
+}
+
+// UseOAuthAuthWithCache provides OAuthAuth authentication for Go SDK with the ability to reuse cached OAuth Token
+// Method is provided as helper to create a default HTTP client that supports OAuth (Service Accounts) authentication.
+// Warning: any previously set httpClient will be overwritten. To fully customize HttpClient use UseHTTPClient method.
```
gssbzn, line 78: just becuase is in a new line doesn't mean it renders in a new line 
```suggestion
//
// Warning: any previously set httpClient will be overwritten. To fully customize HttpClient use UseHTTPClient method.
```

auth/credentials/clientcredentials.go
```diff
@@ -0,0 +1,69 @@
...
+	"net/url"
+	"strings"
+
+	"go.mongodb.org/atlas-sdk/v20241023002/auth"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+// Config describes a 2-legged OAuth2 flow, with both the
+// client application information and the server's endpoint URLs.
+// NOTE: clientcredentials.Config values are used only internally
```
gssbzn, line 18:
```suggestion
//
// NOTE: clientcredentials.Config values are used only internally
```

admin/atlas_client.go
wtrocki: Yes. Makes sense.

auth/credentials/transport.go
```diff
@@ -1,36 +1,38 @@
...
-	return t.UnderlyingTransport.RoundTrip(req)
+	req.Header.Set(userAgent, t.userAgent)
+	return t.underlyingTransport.RoundTrip(req)
 }
 
-// SetAuthHeader sets the Authorization header with the access Token.
-func (t *OAuthCustomHTTPTransport) setAuthHeader(r *http.Request, token *Token) {
-	r.Header.Set("Authorization", "Bearer "+token.AccessToken)
+// NewOAuthCacheTransport http client that adds the User-Agent header and TokenCache
+func NewOAuthCacheTransport(transport http.RoundTripper, userAgent *string, tokenCache *func(token string) error) http.RoundTripper {
```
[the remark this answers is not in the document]
wtrocki: Yes. I have tried to make it part of the admin as It is used in admin to enable admin package specific features we need. 
However admin package is fully generated.

auth/credentials/transport.go
```diff
@@ -1,36 +1,38 @@
...
-// OAuthCustomHTTPTransport is an custom HTTP transport that injects the OAuth2 Token into requests.
-// Transport can be extended by supplying UnderlyingTransport
-// TokenSource can be created by credentials.NewTokenSource
-type OAuthCustomHTTPTransport struct {
-	// UnderlyingTransport is the base RoundTripper used to make HTTP requests.
-	UnderlyingTransport http.RoundTripper
-	// TokenSource used to obtain valid OAuth Token
-	TokenSource TokenSource
+// OAuthTransport provides a way to cache OAuth Token
+type OAuthTransport struct {
```
wtrocki, line 9: Will make this private.

wtrocki:
Applied fixes. I will add Token cache example in separate PR.

wtrocki (commented):
@gssbzn changes LGTM

admin/atlas_client.go
```diff
@@ -55,28 +57,24 @@ func UseDigestAuth(apiKey, apiSecret string) ClientModifier {
 
 // UseOAuthAuth provides OAuthAuth authentication for Go SDK.
```
jeroenvervaeke, line 58:
```suggestion
// UseOAuthAuth provides OAuth authentication for Go SDK.
```

jeroenvervaeke (approved):
One small comment about docs, besides that LGTM

gssbzn (dismissed):

lantoli (dismissed):

gssbzn (approved):

PR #472, opened 2024-11-05
Reorder openapi to make json format always come first based on IPA

wtrocki:
## Description

WIP to showcase impact of the IPA to SDK.

Link to any related issue(s): 

## Type of change:

## Required Checklist:


tools/transformer/src/transformations/reorderAndCleanResponseBodies.js
```diff
@@ -0,0 +1,75 @@
+module.exports = function reorderAndCleanResponseBodies(api) {
+    const hasResponses = api && api.paths;
```
wtrocki, line 2: Can be greatly simplfied if we change it to support latest json type first. 
Since support for multiple responses will require different types I think we can always have accept to list only `application/vnd.atlas.2023-01-01+json` etc.

PR #474, opened 2024-11-07
feat: Preview in a different module path 

lantoli:
## Progress
- Change verified in terraform.


.github/workflows/autoupdate-preview.yaml
```diff
@@ -56,7 +60,20 @@ jobs:
             ## Using early preview changes
 
             ```sh
-            go get go.mongodb.org/atlas-sdk@dev-latest   
+            go get github.com/mongodb/atlas-sdk-go@dev-latest
```
wtrocki, line 63: Kind reminder that we have internal wiki that might need update post merge.

lantoli: yes, I was planning to update wiki once this PR is merged

.github/workflows/autoupdate-preview.yaml
```diff
@@ -56,7 +60,20 @@ jobs:
...
+            > NOTE: You have to run the `go get` command every time you want to update the Preview version to the latest version.
+            
+            After that you can use it as usual, e.g.:
+            ```go
+            ...
+            import(
+              adminPreview "github.com/mongodb/atlas-sdk-go/admin"
+            ...
+              		
+            params := &adminPreview.ListProjectsApiParams{ ...
```
wtrocki, line 75: [Separately to PR] Please note that this will require separate SDK client initialization.

tools/releaser/README.md
```diff
@@ -63,6 +63,11 @@ If the release contains breaking changes (it is a major release)
 we should provide a list of breaking changes notes under `./breaking _changes/{major_version}.md` file.
 This file is automatically generated during release.
 
+## SDK Preview
+
+The [Generate Preview SDK Github action](../../.github/workflows/autoupdate-preview.yaml) creates a PR with latest changes.
+It's created in package `github.com/mongodb/atlas-sdk-go` instead of `go.mongodb.org/atlas-sdk/vXXXXXXXXXXX` so no official release is needed for Preview and it makes it clear that it's not a regular version.
```
wtrocki, line 69:
```suggestion
Preview generates SDK version by substituting`github.com/mongodb/atlas-sdk-go` as package name. 
Version should never be commited to the main branch. This task should only be used in  [Generate Preview SDK Github action](../../.github/workflows/autoupdate-preview.yaml) creates a PR with latest changes.
```

.github/workflows/autoupdate-preview.yaml
lantoli: yes, this was just an example of how to refer to the new Preview module while being able to use the regular one

tools/releaser/scripts/version-paths.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 
 export VERSION_UPDATE_PATHS="../**/go.mod,../**/*.go,../tools/**/*.mustache,../.mockery.yaml,../**/*.md"
+
+# don't change doc files to reduce Preview PR noise
+export VERSION_UPDATE_PATHS_PREVIEW="../**/go.mod,../**/*.go,../tools/**/*.mustache,../.mockery.yaml"
```
wtrocki, line 6: Original file was created to reuse variable across multiple scripts. In this case variable can be moved to original script.

wtrocki (dismissed):
LGTM! Glad to see this change.
I have not verified usability of the generated package.

.github/workflows/autoupdate-preview.yaml
wtrocki: I think you could be better with example/docs as reference rather than code being present in PR description. 
It is size restricted and can lead to confusion

tools/releaser/scripts/preview-version.sh
```diff
@@ -0,0 +1,23 @@
...
+source "$script_path/extract-version.sh"
+# shellcheck source=/dev/null
+source "$script_path/version-paths.sh"
+
+OLD_PACKAGE="go.mongodb.org/atlas-sdk/${SDK_MAJOR_VERSION}"
+NEW_PACKAGE="github.com/mongodb/atlas-sdk-go"
+examples_path="$script_path/../../../examples"
+
+echo "Delete specific version $SDK_MAJOR_VERSION from examples go.mod"
+(cd "$examples_path" && go mod edit -droprequire="$OLD_PACKAGE")
```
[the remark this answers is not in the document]
lantoli: because the replace,  if not done this is the change:
change

and we get this error when running in examples folder go test ./...

```
go: errors parsing go.mod:
go.mod:12:2: require github.com/mongodb/atlas-sdk-go: version "v20241023002.1.0" invalid: should be v0 or v1, not v20241023002
```

lantoli:
> I have not verified usability of the generated package.

@wtrocki I've done it from Terraform.

.github/workflows/autoupdate-preview.yaml
lantoli: that's a good point, will removed from PR description

wtrocki (approved):

tools/releaser/scripts/preview-version.sh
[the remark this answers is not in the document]
wtrocki: Yes. That is what I have expected. As alternative you can discuss with team if we need to run tests for examples or even compile them on the dev version as this will lead to many false compilation issues where examples outdated.
There are pros and cons of this approach thus likely not good to make this change in this PR.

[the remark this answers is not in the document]
lantoli: if we don't do this code-health PR check will fail, I think it's good that the generated PR has all PR checks green

wtrocki (approved):

EspenAlbert (approved):
Nice! I am looking forward to reading the document too 👏

PR #499, opened 2024-12-17
chore: Fixes breaking change log with goreleaser

maastha:
## Description

Fixes breaking change log with goreleaser.

Generated test changlog: https://github.com/mongodb/atlas-sdk-go/blob/89fdc186ed4b2c1c1b3c95559bfab03d3f67a291/tools/releaser/breaking_changes/v20241113004.md

Link to any related issue(s): [CLOUDP-281143](https://jira.mongodb.org/browse/CLOUDP-281143)

## Type of change:

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run `make fmt` and formatted my code

PR #519, opened 2025-02-24
chore: Fixes new-version.sh to avoid accidentally overwriting version variables when calling breaking-changes script

EspenAlbert:
## Description

Fixes new-version.sh to avoid leaking version variables when calling breaking-changes script

Link to any related issue(s): CLOUDP-302378

### Testing
- Ran the generation [here](https://github.com/mongodb/atlas-sdk-go/actions/runs/13515826810/job/37764226741) and it updated `version.go` correctly
  - See the resulting PR opened by the bot in #521 

### More background

- As seen by the final line in `Release Update` [step](https://github.com/mongodb/atlas-sdk-go/actions/runs/13450749746/job/37584642959#step:11:466):
`Creating new version.go file with v20241113005.0.0 and resource version: 20250219`
The `SDK_VERSION` got accidentally overwritten.
This is due to the call to
`source "$script_path/breaking-changes.sh"` that was changed in #499 to `source "$script_path/extract-version.sh"`  leading to overwrite of `SDK_VERSION` to the "old" value from `version.go`

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/core/version.go
+++ b/internal/core/version.go
@@ -5,7 +5,7 @@ package core
 // For more information please see: https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_1_concepts.md
 const (
 	// SDK release tag version.
-	Version = "v20241113005.0.0"
+	Version = "v20250219001.0.0"
 	// Resource Version.
 	Resource = "20250219"
 )
--- a/tools/releaser/scripts/new-version.sh
+++ b/tools/releaser/scripts/new-version.sh
@@ -10,6 +10,17 @@ source "$script_path/extract-version.sh"
 # shellcheck source=/dev/null
 source "$script_path/version-paths.sh"
 
+breakingChanges() {
+	# shellcheck source=/dev/null
+	# Create an isolated subshell to contain the sourcing otherwise SDK_VERSION will be overwritten
+	local breaking_changes=$(
+		source "$script_path/breaking-changes.sh"
+		# Output only the variable we want
+		echo "$BREAKING_CHANGES"
+	)
+	export BREAKING_CHANGES="$breaking_changes"
+}
+
 # Update the version.go file with the new version
 if [ "$NEW_RESOURCE_VERSION" == "$SDK_RESOURCE_VERSION" ]; then
 	echo "Resource Version is already up to date. Changing minor version."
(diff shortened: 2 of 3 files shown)


internal/core/version.go
```diff
@@ -5,7 +5,7 @@ package core
 // For more information please see: https://github.com/mongodb/atlas-sdk-go/blob/main/docs/doc_1_concepts.md
 const (
 	// SDK release tag version.
-	Version = "v20241113005.0.0"
+	Version = "v20250219001.0.0"
```
EspenAlbert, line 8: Manual update as the `new-version.sh` will not enter the major version bump:
`if [ "$NEW_RESOURCE_VERSION" == "$SDK_RESOURCE_VERSION" ];` <-- This will be true and therefore only bump the current, for example: `v20241113005.0.0` --> `v20241113006.0.0`

tools/releaser/scripts/new-version.sh
```diff
@@ -10,6 +10,25 @@ source "$script_path/extract-version.sh"
 # shellcheck source=/dev/null
 source "$script_path/version-paths.sh"
 
+breakingChanges() {
```
lantoli, line 13: I think that I would be better that `breaking-changes.sh` has the logic to not override env.var. so it can be called from any script, instead of putting the responsibility in the script caller, e.g. if it needs to be called from another script we'll need something like this.    

So instead, the script could save the env.var,   do its things, and the get back the env.vars. 

quick example (to review):
```
# Save environment variables to a temporary file
export -p > /tmp/saved_env.sh

# Do your work here that might change environment variables
export PATH=/custom/path
# ... other operations ...

# Restore the original environment
source /tmp/saved_env.sh

# Optional: clean up
rm /tmp/saved_env.sh
```

wtrocki: Alternatively consider python/node.js script as that bash specifically does a lot non typical bash work.

wtrocki (approved):

internal/core/version.go
wtrocki: Consider bumping in the separate PR so we can test this with dev version first

tools/releaser/scripts/new-version.sh
EspenAlbert: Thanks. Changed in c9ad7be17d876c0d2a0c871b65bc2225ac360355

lantoli (approved):
LGTM, thanks for the change to move the logic to the script

oarbusi (approved):

PR #534, opened 2025-03-14
task: update dependabot for npm

gssbzn:
## Description

dependabot seems to be not working for NPM trying to pint it to the folder, also updating the reviewer to integrations

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,11 +17,11 @@ updates:
       interval: monthly
       day: monday
     reviewers:
-      - "mongodb/APIx1"
+      - "mongodb/APIx-Integrations"
   - package-ecosystem: npm
-    directory: "/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -17,11 +17,11 @@ updates:
       interval: monthly
       day: monday
     reviewers:
-      - "mongodb/APIx1"
+      - "mongodb/APIx-Integrations"
```
fmenezes, line 20: shouldn't CODEOWNERS catch this? perhaps we don't need it

fmenezes (approved):

.github/dependabot.yml
gssbzn: yeah I was tempted to remove it to be honest so I can do it

PR #547, opened 2025-03-21
CLOUDP-307850: remove required from merged field

wtrocki:
## Description

Example on how to remove required field from merged objects
Screenshot 2025-03-21 at 11 59 32

### DIFF
--- a/admin/model_search_index_create_request.go
+++ b/admin/model_search_index_create_request.go
@@ -5,11 +5,11 @@ package admin
 // SearchIndexCreateRequest struct for SearchIndexCreateRequest
 type SearchIndexCreateRequest struct {
 	// Label that identifies the collection to create an Atlas Search index in.
-	CollectionName *string `json:"collectionName,omitempty"`
+	CollectionName string `json:"collectionName"`
 	// Label that identifies the database that contains the collection to create an Atlas Search index in.
-	Database *string `json:"database,omitempty"`
+	Database string `json:"database"`
 	// Label that identifies this index. Within each namespace, names of all indexes in the namespace must be unique.
-	Name *string `json:"name,omitempty"`
+	Name string `json:"name"`
 	// Type of the index. The default type is search.
 	Type       *string                                 `json:"type,omitempty"`
 	Definition *BaseSearchIndexCreateRequestDefinition `json:"definition,omitempty"`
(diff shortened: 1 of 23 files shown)


oarbusi (approved):

admin/model_search_index_create_request.go
```diff
@@ -5,11 +5,11 @@ package admin
...
 	// Label that identifies the collection to create an Atlas Search index in.
-	CollectionName string `json:"collectionName"`
+	CollectionName *string `json:"collectionName,omitempty"`
 	// Label that identifies the database that contains the collection to create an Atlas Search index in.
-	Database string `json:"database"`
+	Database *string `json:"database,omitempty"`
 	// Label that identifies this index. Within each namespace, names of all indexes in the namespace must be unique.
-	Name string `json:"name"`
+	Name *string `json:"name,omitempty"`
 	// Type of the index. The default type is search.
```
AgustinBettati, line 13: From my understanding this is removing required from all properties present in oneOf schemas. 
This does fix issue in CLOUDP-307850 but is also impacting broader cases were this transformation from required to optional is not needed. 
Selected this specific case of search index, `collectionName` `database` and `name` are required properties in all the oneOf schemas (vector and text search) so we should have no need of marking it as optional.

Would be great if we can dedicate some more time to get a more specific fix.

EspenAlbert (approved):
An alternative method is to keep the struct changes as is here. But add more helper methods to construct the structs where each helper method has the required fields for that oneof.

admin/model_search_index_create_request.go
wtrocki: Agree. We can make logic to only remove it when not all types have it required etc.

wtrocki:
> An alternative method is to keep the struct changes as is here. But add more helper methods to construct the structs where each helper method has the required fields for that oneof.

Requires change in parser and will lead to invisible breaking changes (in parser) for customers.

wtrocki:
Closing to make it clear that PR is just example.

tools/transformer/src/transformations/oneOf.js
```diff
@@ -121,6 +129,24 @@ function transformOneOfProperties(parentObject, api) {
       ...parentObject.properties,
       ...childProperties,
     };
+
+    // Collect required fields from the child
+    if (childObject.required) {
```
wtrocki, line 134: This logic should give you some context on complexity of that transformation.
Results seem off - will likely need more investigation.

tools/transformer/src/transformations/oneOf.js
```diff
@@ -121,6 +129,24 @@ function transformOneOfProperties(parentObject, api) {
...
+      allRequiredFields.push(childObject.required);
+    }
+  }
+
+  // Handle required properties
+  if (allRequiredFields.length > 0) {
+    const commonRequired = allRequiredFields.reduce((common, current) =>
```
wtrocki, line 141: We can use set instead

PR #555, opened 2025-03-27
CLOUDP-307850: merge oneOf required

wtrocki:
## Description

Refined example showcasing all possible changes for alternative, where we merge only required fields that exist on every child and parent. 

See https://github.com/mongodb/atlas-sdk-go/pull/547

### DIFF
--- a/openapi/atlas-api-transformed.yaml
+++ b/openapi/atlas-api-transformed.yaml
@@ -731,6 +731,13 @@ components:
           example: STARTED
           readOnly: true
           type: string
+      required:
+        - alertConfigId
+        - created
+        - eventTypeName
+        - id
+        - status
+        - updated
     AlertsNotificationRootForGroup:
       description: One target that MongoDB Cloud sends notifications when an alert triggers.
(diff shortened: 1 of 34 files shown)


openapi/atlas-api-transformed.yaml
```diff
@@ -731,6 +731,13 @@ components:
           example: STARTED
           readOnly: true
           type: string
+      required:
```
wtrocki, line 734: Requires OpenAPI fix as not all of those look correct

openapi/atlas-api-transformed.yaml
```diff
@@ -5306,8 +5315,6 @@ components:
             compatibility for deprecated provider types and/or instance sizes.
           readOnly: true
           type: string
-      required:
```
wtrocki, line 5309: Discriminator field should be required. Lack of presence of requirement on some subfields usually means missconfiguration

oarbusi:
Won't be merging this in the short term. Closed for now

PR #557, opened 2025-03-28
fix: Makes specified fields optional in their respective schemas

oarbusi:
### Problem description:
Atlas Go SDK was autogenerated and included this [change](https://github.com/mongodb/atlas-sdk-go/pull/545/files#diff-05814b892b574662623416b5fa87eece63ecf7a2e7b5abf8bdafaedbe9fbf414R35) in the release [v20250312](https://github.com/mongodb/atlas-sdk-go/releases/tag/v20250312001.0.0), which removed GreaterThanRawThreshold and replaced it with StreamProcessorMetricThreshold.

In the [OpenAPI spec](https://raw.githubusercontent.com/mongodb/atlas-sdk-go/refs/heads/main/openapi/atlas-api.yaml), GroupAlertsConfig is defined as OneOf of multiple schemas. The latest one added, StreamProcessorMetricAlertConfigViewForNdsGroup has caused a change in the [transformed OpenAPI spec](https://github.com/mongodb/atlas-sdk-go/blob/92ba675c039b5e9427084c208fc35f957dee2385/openapi/atlas-api-transformed.yaml#L9869) of the threshold property changing from to  GreaterThanRawThreshold to [StreamProcessorMetricThreshold](https://github.com/mongodb/atlas-sdk-go/blob/92ba675c039b5e9427084c208fc35f957dee2385/openapi/atlas-api-transformed.yaml#L15329) which defines metricName as required, and in the SDK is a string(instead of a pointer). This causes previously successful API calls(through the SDK in TF) to fail because metricName is included in the request as an empty string (full API calls in the attached file).

### Fix:
Makes metricName field optional in their respective schema

Link to any related issue(s): CLOUDP-307850

## Type of change:

## Required Checklist:

### DIFF
--- a/openapi/atlas-api-transformed.yaml
+++ b/openapi/atlas-api-transformed.yaml
@@ -15349,8 +15349,6 @@ components:
           $ref: "#/components/schemas/RawMetricUnits"
           description: Element used to express the quantity. This can be an element of
             time, storage capacity, and the like.
-      required:
-        - metricName
       title: Stream Processor Metric Threshold
       type: object
     StreamsAWSConnectionConfig:
--- a/tools/transformer/src/field.transformations.json
+++ b/tools/transformer/src/field.transformations.json
@@ -1,0 +1,7 @@
+{
+	"optionalFields": {
+		"StreamProcessorMetricThreshold": [
+			"metricName"
+		]
+	}
+}
--- a/tools/transformer/src/field.transformations.json
+++ b/tools/transformer/src/field.transformations.json
@@ -1,7 +1,5 @@
 {
-	"optionalFields": {
-		"StreamProcessorMetricThreshold": [
-			"metricName"
-		]
-	}
+  "optionalFields": {
+    "StreamProcessorMetricThreshold": ["metricName"]
+  }
 }
--- a/tools/transformer/src/transformations/fields.js
+++ b/tools/transformer/src/transformations/fields.js
@@ -6,22 +6,24 @@ const fieldTransformations = require("../field.transformations.json");
  * @returns OpenAPI JSON File
  */
 function applyFieldTransformations(api) {
-	const { optionalFields } = fieldTransformations;
+  const { optionalFields } = fieldTransformations;
 
-	// Make specified fields optional in their schemas
-	for (const [schemaName, fields] of Object.entries(optionalFields)) {
-		const schema = api.components.schemas[schemaName];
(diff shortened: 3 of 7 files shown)


tools/transformer/src/field.transformations.json
```diff
@@ -0,0 +1,5 @@
+{
```
marcosuma, line 1: n00b question: what are these? are they auto-generated?

marcosuma (approved):
LGTM - good to have another pair of eyes from the team - thank you for coordinating this!

tools/transformer/src/field.transformations.json
oarbusi: We create these json ourselves. Like [duplicate.ignore.json](https://github.com/mongodb/atlas-sdk-go/blob/main/tools/transformer/src/duplicate.ignore.json).
In this case I defined it in a way so that I can map the schema<>field so that then I use it in `tools/transformer/src/transformations/fields.js` to make metricName optional.

openapi/atlas-api-transformed.yaml
```diff
@@ -15349,8 +15349,6 @@ components:
           $ref: "#/components/schemas/RawMetricUnits"
           description: Element used to express the quantity. This can be an element of
             time, storage capacity, and the like.
-      required:
```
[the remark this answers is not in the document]
oarbusi: IMO in the code we already capture it in the custom logic we are adding to remove this required. Or maybe you are referring to adding this to changelog or something similar?

tools/transformer/src/transformations/fields.js
```diff
@@ -0,0 +1,29 @@
+const fieldTransformations = require("../field.transformations.json");
+
+/**
+ * Makes specified fields optional in their respective schemas
+ * @param {*} api OpenAPI JSON File
+ * @returns OpenAPI JSON File
+ */
+function applyFieldTransformations(api) {
+  const { optionalFields } = fieldTransformations;
```
AgustinBettati, line 9: find it strange why we add this level of indirection with the file, instead of just having the array here

AgustinBettati (approved):
great that we get a release with the fix, nice approach to give us time

EspenAlbert (approved):

maastha:
@oarbusi I think we need to update this https://github.com/mongodb/atlas-sdk-go/blob/main/tools/transformer/README.md

PR #252, opened 2024-01-04
CLOUDP-217493 - Arrays as pointers

wtrocki:
## Description

Refactor to make SDK use pointers for non-required arrays.  

Working on docs.

## Progress

- [x] Template change
- [x] Regenerated code

## Follow up PR 
- [ ] Documentation updates
- [ ] Usage in Atlas CLI
- [ ] Usage in CDK
Link to any related issue(s): CLOUDP-217493

## Testing

Run all the examples locally

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #779, opened 2026-04-27
chore: Add pointer to optional free-form map fields in generated models

lantoli:
## Description

Optional fields typed as `map[string]any` (generated from OpenAPI schemas with `additionalProperties: true`) were missing the pointer that optional array fields (`*[]T`) have had since #252. This meant callers could not distinguish between an unset map and an empty map, and setters had no way to express "clear this field."

This change updates the mustache template (`tools/config/go-templates/model_simple.mustache`) so that optional free-form map fields use `*map[string]any` instead of `map[string]any`. Non-map free-form fields (typed as `any`) are unaffected. This is not a breaking change: no existing field in the SDK currently matches this pattern, so no generated Go code changes on `main`. The pointer will apply to new `map[string]any` fields as they are introduced, specifically the `IntelligentWorkloadManagementPolicyOverrides` field being added as part of the IWM project (CLOUDP-387563).

The fix mirrors the approach used for array fields and is consistent with how maps backed by a typed schema (e.g. `SearchMappings.Fields`) already generate `*map[string]T`.

**Template note:** The mustache template uses `{{^isFreeFormObject}}*{{/isFreeFormObject}}{{#isFreeFormObject}}{{#isMap}}*{{/isMap}}{{/isFreeFormObject}}` — the verbose form is necessary because Mustache has no OR operator. It emits `*` for all optional fields except free-form non-map types (`any`), which must remain unpointed since they have no nil sentinel.

Link to any related issue(s): CLOUDP-400281

## Type of change:

## Required Checklist:

## Further comments

### DIFF
--- a/tools/config/go-templates/model_simple.mustache
+++ b/tools/config/go-templates/model_simple.mustache
@@ -26,7 +26,7 @@ type {{classname}} struct {
 {{#deprecated}}
 	// Deprecated
 {{/deprecated}}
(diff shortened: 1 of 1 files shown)


AgustinBettati (approved):
LGTM, thanks for the clear PR description

PR #823, opened 2026-07-06
feat: Add preview media-type support to the Preview SDK

oarbusi:
## Problem

SDK generation drops any API or field exposed only via
`application/vnd.atlas.preview+json`, so features moved behind the preview media
type disappear from the SDK and block Terraform provider development against them.

The existing Preview SDK (`dev-latest`, `go get github.com/mongodb/atlas-sdk-go@dev-latest`)
targets the latest stable dated version and filters out `preview`, so
preview-header endpoints and fields were excluded there too.

## Solution

Opt-in via a new `TARGET_API_VERSION=preview` env var (production pipeline
unaffected). When set, the pipeline fetches the latest stable dated spec as the
base, fetches `openapi-preview.yaml` as a delta, and merges the delta onto the
base before generation. The result is a superset SDK: every stable endpoint plus
the preview media-type ones. No new import path and no downstream alias —
consumers keep using `@dev-latest`.

`openapi-preview.yaml` is a delta (~20 operations), not a full snapshot, so it
must be merged rather than used on its own.

Merge granularity:
- Same path + method in both → preview overrides (one method, preview shape).
- Same path, method only in preview → both kept.
- New preview path/operation → added as-is.

Preview override is a deliberate trade-off: OpenAPI cannot hold two operations for
one path+method and the generator emits one method per operation, so exposing both
variants would require template changes shared with the production SDK. For a
development-only SDK this is correct. Stable-only endpoints are untouched.

The same whole-object override applies to `components.schemas`: when the preview
delta and the stable base declare a schema with the same name, the preview
definition replaces the base one entirely (not a field-level merge).

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s): CLOUDP-417642

## Type of change:

## Required Checklist:


tools/scripts/fetch.sh
```diff
@@ -45,4 +54,28 @@ echo "Fetching api from $openapi_url to $OPENAPI_FILE_NAME"
...
+  # Merge the preview delta onto the stable base spec (in place), producing a
+  # superset spec for the rest of the generation pipeline. Mirrors the
+  # transformer invocation used in transform.sh.
+  echo "Merging preview delta into $OPENAPI_FILE_NAME"
+  npm install
+  npm exec --prefix .. atlas-openapi-transformer -- merge-preview \
+    --base "${OPENAPI_FOLDER}/${OPENAPI_FILE_NAME}" \
+    --preview "${OPENAPI_FOLDER}/atlas-api-preview.yaml" \
+    --output "${OPENAPI_FOLDER}/${OPENAPI_FILE_NAME}"
+  rm -f "${OPENAPI_FOLDER}/atlas-api-preview.yaml"
```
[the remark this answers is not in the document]
oarbusi: addressed in [a94dab4](https://github.com/mongodb/atlas-sdk-go/pull/823/commits/a94dab47d022f51a923502ddedfb2ecfd499eb90)

tools/transformer/src/engine/mergePreview.js
```diff
@@ -0,0 +1,66 @@
...
+          ...entries,
+        };
+      }
+    }
+  }
+
+  // Append preview tags that are not already declared.
+  if (Array.isArray(preview.tags)) {
+    base.tags = base.tags || [];
+    const existing = new Set(base.tags.map((t) => t && t.name));
```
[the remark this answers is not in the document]
oarbusi: if preview.tags itself has no duplicates (which it doesn't), the current code is correct.

.github/workflows/autoupdate-preview.yaml
```diff
@@ -62,6 +63,12 @@ jobs:
             Automatic update for MongoDB Atlas Go Client based on *release candidate* OpenAPI file.
             **DO NOT MERGE THIS PR TO THE MAIN BRANCH**
 
+            > [!WARNING]
```
AgustinBettati, line 67: Would take the opportunity to be very explicit on how this Preview version is being generated. As per the PR body we could mention: 

> This preview SDK is not officially supported and is not production ready. It is generated from the `dev` OpenAPI spec and additionally merges the preview delta (`openapi-preview.yaml`, media type `application/vnd.atlas.preview+json`) on top of it

And for the PR title we could go for something like `APIBot: Dev Preview SDK (dev spec + preview APIs) — DO NOT MERGE`

Noting as `GO SDK Dev Preview` PR title has caused confusion in the past.

tools/transformer/src/engine/mergePreview.js
```diff
@@ -0,0 +1,66 @@
+/**
```
AgustinBettati, line 1: One alternative to avoid any form of merging could be to have 2 separate PRs, one generating with cloud-dev API Spec, another generating with cloud-dev preview API Spec.
Merge logic appears trivial so okay to leave as is.

AgustinBettati (approved):
LGTM

.github/workflows/autoupdate-preview.yaml
oarbusi: good point, changed the generated PR title and description following your suggestion in [dc628c1](https://github.com/mongodb/atlas-sdk-go/pull/823/commits/dc628c1fd4ef4807a67ae5bf97a34298289a19d4)

tools/transformer/src/engine/mergePreview.js
oarbusi: because as you said, merge logic is simple, I prefer to keep it as is so we keep the number of PRs that generate different versions of the SDK to a minimum

.github/workflows/autoupdate-preview.yaml
```diff
@@ -54,14 +55,23 @@ jobs:
         if: env.FILES_CHANGED == 'true'
         with:
           token: ${{ secrets.APIX_BOT_PAT }}
-          title: "APIBot: GO SDK Dev Preview"
+          title: "APIBot: Dev Preview SDK (dev spec + preview APIs) — DO NOT MERGE"
```
AgustinBettati, line 58: nit: We could also adjust the PR to be opened as a draft right? Making more explicit intention is not for it to be merged, more to just showcase changes and describe how to consume the generated result.

tools/transformer/__tests__/mergePreview.test.js
```diff
@@ -0,0 +1,127 @@
...
+      },
+      parameters: {
+        previewParam: { name: "preview", in: "query" },
+      },
+    },
+  };
+
+  const merged = mergePreview(baseDoc(), preview);
+
+  expect(merged.components.schemas.Cluster.properties).toHaveProperty(
```
maastha, line 113: what about the cluster id field? I think that should still exist right?

tools/transformer/src/engine/mergePreview.js
```diff
@@ -0,0 +1,66 @@
...
+ * The published `openapi-preview.yaml` only contains the operations that have
+ * preview content (a small delta), not the full API surface. To produce a
+ * usable preview SDK we overlay that delta on top of the latest stable spec so
+ * the result is a superset: every stable endpoint plus the preview ones.
+ *
+ * Merge rules:
+ *  - paths: preview paths/methods are added; a method present in both is
+ *    overridden by the preview version (so preview media types win).
+ *  - components.*: preview entries (schemas, parameters, responses,
+ *    requestBodies, headers, securitySchemes, ...) are added/overridden so that
```
maastha: Wanted to check my understanding of the merge here — when a preview entry and a base entry share a name, does the preview one replace the whole thing, or just the fields that overlap?  Is openapi-preview.yaml guaranteed to always spell out the complete object for anything it re-declares?

.github/workflows/autoupdate-preview.yaml
oarbusi: good point, this is an easy change. Done in [de6cd14](https://github.com/mongodb/atlas-sdk-go/pull/823/commits/de6cd143c129ba8c448bdec1f3c0cf48ad5cced7)

tools/transformer/__tests__/mergePreview.test.js
oarbusi: yes sorry. Fixed in [a479a6e](https://github.com/mongodb/atlas-sdk-go/pull/823/commits/a479a6e5c85f232e8e6e158e5ff97c03b205e914)

tools/transformer/src/engine/mergePreview.js
oarbusi: Correct, when a name collides, the preview entry replaces the whole thing (whole-object override, not a field-level merge).

I checked preview against the current dev spec: all 31 schema name collisions between the preview delta and the base spec redeclare the complete object, none are partial.

If that ever changed (a preview entry redeclaring only some fields), we'd silently lose the missing base fields for that entry. Documented this behavior + the test verifying it in mergePreview.js/mergePreview.test.js, so it's visible if it needs revisiting.

oarbusi:
generated PR after running the GHA from this branch https://github.com/mongodb/atlas-sdk-go/pull/824

lantoli (approved):

PR #825, opened 2026-07-09
feat: Includes private preview APIs in preview SDK

oarbusi:
## Description

Extended the preview-SDK pipeline from https://github.com/mongodb/atlas-sdk-go/pull/823 to also merge private preview specs, not just the public one.

Private preview isn't one delta file like `openapi-preview.yaml`, it's a directory (`openapi/v2/private/`) of per-feature delta files (rate-limit, sandbox-resource, etc.), with its own versions.json listing which ones are currently active. Each file has the same paths/components/tags delta shape as the public preview spec, so the existing mergePreview overlay logic applies unchanged, it just needed to be applied once per delta file instead of once.

Link to any related issue(s): CLOUDP-417642

## Type of change:

## Required Checklist:


tools/transformer/src/cli.js
```diff
@@ -74,21 +74,31 @@ yargs(hideBin(process.argv))
...
+            "(e.g. the public preview spec plus one or more private preview " +
+            "specs) on top of the base, applied in the order given.",
+        })
         .option("output", { alias: "o", type: "string", demandOption: true })
         .option("log-level", { type: "string" }),
     async (args) => {
       const log = configureLogger(args["log-level"]);
       const originalConsole = global.console;
       global.console = log;
       try {
```
lantoli, line 96: could we add a CLI-level test with two `--preview` flags? The current test verifies mergePreview composition, but it would still pass if the yargs array parsing or CLI iteration regressed

lantoli (approved):

tools/transformer/src/cli.js
oarbusi: added CLI-level tests in [a16f015](https://github.com/mongodb/atlas-sdk-go/pull/825/commits/a16f0153e953c1c6334037644e55e64d311b5e6b)

oarbusi:
CI fails because SDK generation changes are not in main yet. Will be fixed once new release is done

PR #827, opened 2026-07-09
feat: Adds NullFields field to be able to send nulls

oarbusi:
## Description

Adds a `NullFields` mechanism to the Go client so scalar (and other) fields can be sent as explicit JSON `null` in request bodies: with `omitempty`, a field can be *omitted* or *have a value*, but never *explicitly null*, which some APIs require to clear/unset a field (e.g. `waveAssignment`, see https://github.com/mongodb/atlas-sdk-go/pull/801.

Usage: `obj.SetWaveAssignmentNil()` — everything else is unchanged and non-breaking.

applies to plain/flat models only; `oneOf`/`anyOf` composites and array-wrapper models already have their own custom marshaling and are unaffected.

Link to any related issue(s): CLOUDP-315290

## Type of change:

## Required Checklist:

## Manual testing

Tested the regenerated SDK from a local checkout of this branch using a small standalone Go program (not part of this repo), covering two layers:

1. **Direct marshaling** — confirms `SetXxxNil()` + the generated `MarshalJSON` override correctly force a `null`, even overriding a value set immediately before.
2. **End-to-end through the real generated client** — spun up an `httptest.Server` and called the actual generated `ProjectIPAccessListApi.CreateAccessListEntry(...).Execute()` method to capture the literal HTTP request body sent over the wire, to rule out any difference between `json.Marshal` and the client's internal encoding path (`setBody` → `json.NewEncoder(...).Encode(...)`).

### Test code

```go
package main

import (
	"context"
	"encoding/json"
	"fmt"
	"io"
	"net/http"
	"net/http/httptest"

	"go.mongodb.org/atlas-sdk/v20250312021/admin"
)

func main() {
	fmt.Println("=== 1) Direct json.Marshal of the model ===")
	item := admin.NewNetworkPermissionEntry()
	item.SetIpAddress("1.2.3.4")
	item.SetComment("temporary access")
	item.SetCommentNil() // override with explicit null
	b, _ := json.Marshal(item)
	fmt.Println(string(b))

	fmt.Println()
	fmt.Println("=== 2) Actual HTTP body sent by the generated API client ===")

	var capturedBody []byte
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		capturedBody, _ = io.ReadAll(r.Body)
		w.Header().Set("Content-Type", "application/json")
		w.WriteHeader(http.StatusCreated)
		_, _ = w.Write([]byte(`[]`))
	}))
	defer server.Close()

	cfg := admin.NewConfiguration()
	cfg.Servers = admin.ServerConfigurations{{URL: server.URL}}
	client := admin.NewAPIClient(cfg)

	entry := admin.NewNetworkPermissionEntry()
	entry.SetIpAddress("1.2.3.4")
	entry.SetComment("temporary access")
	entry.SetCommentNil() // force explicit null over the wire

	entries := []admin.NetworkPermissionEntry{*entry}
	_, _, err := client.ProjectIPAccessListApi.CreateAccessListEntry(context.Background(), "000000000000000000000000", &entries).Execute()
	if err != nil {
		fmt.Println("request error (expected, response mock body is empty):", err)
	}
	fmt.Println("raw request body received by server:")
	fmt.Println(string(capturedBody))
}
```

### Output

```
=== 1) Direct json.Marshal of the model ===
{"comment":null,"ipAddress":"1.2.3.4"}

=== 2) Actual HTTP body sent by the generated API client ===
request error (expected, response mock body is empty): http: read on closed response body
raw request body received by server:
[{"comment":null,"ipAddress":"1.2.3.4"}]
```

`SetCommentNil()` correctly overrides a previously-set value (`"temporary access"`) and produces an explicit JSON `null`, rather than omitting the field.


admin/utils.go
```diff
@@ -94,3 +97,55 @@ func Int64PtrToIntPtr(i64 *int64) *int {
...
+			continue
+		}
+		name, _, _ := strings.Cut(tag, ",")
+		tags[field.Name] = name
+	}
+	jsonTagCache.Store(t, tags)
+	return tags
+}
+
+// marshalWithNullFields marshals v to JSON as usual, then patches the result so that fields named
```
lantoli: consider adding unit tests for `marshalWithNullFields` (empty nullFields, single/multiple null fields, unknown field name) as it does a double marshal/unmarshal with reflection

admin/model_access_list_item.go
```diff
@@ -10,6 +10,15 @@ type AccessListItem struct {
 	// IP address included in the API access list.
 	// Read only field.
 	IpAddress string `json:"ipAddress"`
+	// NullFields is a list of field names (e.g. "FieldName") to send as an explicit JSON null,
+	// overriding the field's actual value.
+	NullFields []string `json:"-"`
```
lantoli: the generated `admin/model_*.go` and `docs/docs/*.md` files should be removed from this PR, they will be regenerated automatically in the SDK publish PR. this PR should only contain the template changes in `tools/config/go-templates/` and the `admin/utils.go` utility

tools/config/go-templates/model_simple.mustache
```diff
@@ -31,12 +31,29 @@ type {{classname}} struct {
 {{#isAdditionalPropertiesTrue}}
 	AdditionalProperties map[string]any
 {{/isAdditionalPropertiesTrue}}
+{{^isArray}}
+{{^isAdditionalPropertiesTrue}}
+	// NullFields is a list of field names (e.g. "FieldName") to send as an explicit JSON null,
+	// overriding the field's actual value.
```
lantoli: also update `model_doc.mustache` to document `SetXxxNil()` and `NullFields` in the generated per-model docs

tools/config/go-templates/model_simple.mustache
```diff
@@ -175,6 +192,16 @@ func (o *{{classname}}) Set{{name}}(v {{vendorExtensions.x-go-base-type}}) {
 	o.{{name}} = {{^isFreeFormObject}}&{{/isFreeFormObject}}{{#isFreeFormObject}}{{#isMap}}&{{/isMap}}{{/isFreeFormObject}}v
 }
 
+{{^isArray}}
+{{^isAdditionalPropertiesTrue}}
+// Set{{name}}Nil sets {{name}} to an explicit JSON null when marshaled.
+func (o *{{classname}}) Set{{name}}Nil() {
+	o.{{name}} = nil
```
lantoli, line 201: q: what happens if `SetXxxNil()` is called and then later `SetXxx(value)`? the field name stays in `NullFields` so the explicit null would override the value, is that intentional?

admin/model_access_list_item.go
oarbusi: CI fails without including changes from generating the SDK https://github.com/mongodb/atlas-sdk-go/actions/runs/29027855901/job/86152830957

lantoli: uhm, i don't think it should fail in this case, is it a required PR check?

admin/utils.go
oarbusi: Good catch, added unit tests in test/utils_test.go covering empty NullFields, single/multiple null fields, and an unknown field name (appended directly to the exported NullFields slice, since it's the only way to reach that branch black-box). All exercised through the real generated MarshalJSON/SetXxxNil methods rather than testing the unexported marshalWithNullFields directly.

tools/config/go-templates/model_simple.mustache
```diff
@@ -31,12 +31,29 @@ type {{classname}} struct {
 {{#isAdditionalPropertiesTrue}}
 	AdditionalProperties map[string]any
 {{/isAdditionalPropertiesTrue}}
+{{^isArray}}
+{{^isAdditionalPropertiesTrue}}
+	// NullFields is a list of field names (e.g. "FieldName") to send as an explicit JSON null,
```
AgustinBettati: nit: would clarify this is an _internal_ field that is never sent as part of the payload (making more clear the ``json:"-"`` behaviour)

tools/config/go-templates/model_simple.mustache
```diff
@@ -31,12 +31,29 @@ type {{classname}} struct {
...
+	// overriding the field's actual value.
+	NullFields []string `json:"-"`
```
AgustinBettati, line 38: q: For what fields types is `SetXxxNil()` defined? Should we clarify this aspect in our docs?

AgustinBettati (commented):
Liking the approach 💯 have some questions

tools/config/go-templates/model_simple.mustache
oarbusi: Good catch, that wasn't intentional. Fixed: Set{{name}}(v) now also calls removeNullField(o.NullFields, "{{name}}") so calling the regular setter after SetXxxNil() clears the null override. Also made SetXxxNil() dedupe via addNullField so calling it twice doesn't grow NullFields. Covered by two new tests in test/utils_test.go: "setting a value after SetXxxNil clears the null override" and "calling SetXxxNil twice does not duplicate the field name".

tools/config/go-templates/model_simple.mustache
oarbusi: Done, added a SetXxxNil section to model_doc.mustache, so the generated per-model docs now document it

admin/model_access_list_item.go
oarbusi: reverted the generation so we have a cleaner PR. Even if CI fails, I think it's okay in this specific case

oarbusi: Included some more unit tests and a required check fails because we are not generating the SDK. Will leave the PR as is until we finish the review and before merging I will generate it to so that CI passes and we don't have to force the merge

tools/config/go-templates/model_simple.mustache
oarbusi: Good point, this actually uncovered a real bug: SetXxxNil() was only being generated for scalar/object optional fields, never for array-typed ones. Fixed so SetXxxNil() is now generated for optional fields of any type, including arrays. Added a test for it in test/utils_test.go.

tools/config/go-templates/model_simple.mustache
oarbusi: Good nit, updated the comment to spell out that NullFields is internal and never sent as part of the payload.

wtrocki:
Do we plan to add new methods to documentation?

oarbusi:
> Do we plan to add new methods to documentation?

@wtrocki not sure I understand the question, could you clarify?

wtrocki:
@oarbusi I'm looking to see if you plan to add any [user facing documentation ](https://github.com/mongodb/atlas-sdk-go/tree/main/docs)

oarbusi:
@wtrocki thanks for clarifying. No, not planned to include this new method to https://github.com/mongodb/atlas-sdk-go/tree/main/docs because it will already be documented [here](https://github.com/mongodb/atlas-sdk-go/tree/main/docs/docs) once we regenerate the SDK

lantoli:
@oarbusi @wtrocki I think it would be useful to add a short customer-facing doc (for example in docs/doc_5_best_practices.md) explaining the context for the new SetXxxNil() methods. The SDK normally omits nil fields because of omitempty, but in a PATCH request you may want to explicitly reset a field by sending null. In that case, use SetXxxNil().

AI assisted doc example:

Add this section right after ## Working with Arrays:
## Working with Explicit Null Values

By default, the Atlas Go SDK omits `nil` fields from request payloads because model fields use `omitempty`. This preserves the existing value on the server and is the right behavior for most requests.

For `PATCH` or other update operations, however, you may want to explicitly reset a field by sending `null`. Call `SetXxxNil()` on the request model to force that field to be sent as `null`:

```go
// Surrounding code omitted for brevity

entry := admin.NewNetworkPermissionEntry()
entry.SetIpAddress("1.2.3.4")
entry.SetCommentNil() // sends "comment": null in the JSON payload

File: test/utils_test.go [hunk 7]
```diff
@@ -102,3 +103,125 @@ func TestStringToTime(t *testing.T) {
...
+		got := unmarshalToMap(t, b)
+		if got["ipAddress"] != "1.2.3.4" {
+			t.Errorf("want ipAddress %q, got %v", "1.2.3.4", got["ipAddress"])
+		}
+		if got["comment"] != "a comment" {
+			t.Errorf("want comment %q, got %v", "a comment", got["comment"])
+		}
+	})
+
+	t.Run("single null field overrides its value", func(t *testing.T) {
```
lantoli: the tests in test/utils_test.go fail because the generated admin/ files aren't included in this PR.
one approach is to add them in a follow-up PR once the release with the new SetXxxNil() methods is published, so they run against real generated code and catch regressions

oarbusi:
Added customer facing doc in [588e34e](https://github.com/mongodb/atlas-sdk-go/pull/827/commits/588e34ec27b03eea4588351edf6841479e9fdf4f) as suggested by @lantoli and @wtrocki

test/utils_test.go
oarbusi: I was planning to generate the SDK here before merging but your option is also viable. Done in [a1d66dc](https://github.com/mongodb/atlas-sdk-go/pull/827/commits/a1d66dcc6c657ce4a34e84095d685553b0582cd0)

lantoli (approved):
LGTM, thanks for the follow-ups

oarbusi:
Merging. CI check that fails is because the SDK has not been regenerated in this PR, it will be done automatically once automation is triggered

PR #838, opened 2026-08-06
chore: Decouple NewConfiguration default URL from spec servers URL

oarbusi:
## Description

`admin.NewConfiguration()` bakes the spec's servers URL into generated code at generation time via `configuration.mustache`. If that URL were to ever change in the spec, regenerating the SDK would silently change the default base URL for consumers on the `NewConfiguration()` + `NewAPIClient` path, a behavior change that could be treated as breaking and force a major version bump.

This PR adjusts `configuration.mustache` so `NewConfiguration()` reuses the existing hand-written `core.DefaultCloudURL` constant (already used by the recommended admin.NewClient path) instead of the spec's servers URL. Generation is now decoupled from the spec servers block, so any future spec URL change will not alter generated code.

## Verification:
Regenerated the SDK against a spec copy with a different servers URL and confirmed the default base URL in generated code does not change. `go build ./...` and `go test ./test` pass.

Link to any related issue(s): CLOUDP-422448

## Type of change:

## Required Checklist:

## Further comments

No automated test was added: the fix lives in the generation template and was verified by regenerating against a spec with a different servers URL (produces no diff in the generated default base URL). Consumers overriding the base URL (UseBaseURL, or mutating cfg.Servers directly, e.g. for `cloud-dev.mongodb.com`) are unaffected.

### DIFF
--- a/admin/configuration.go
+++ b/admin/configuration.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+
+	"go.mongodb.org/atlas-sdk/v20250312023/internal/core"
 )
 
 // contextKeys are used to identify the type of value in the context.
@@ -82,7 +84,8 @@ func NewConfiguration() *Configuration {
 		Debug:         false,
 		Servers: ServerConfigurations{
 			{
-				URL:         "https://cloud.mongodb.com",
+				// Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code.
+				URL:         core.DefaultCloudURL,
 				Description: "No description provided",
 			},
 		},
--- a/admin/configuration.go
+++ b/admin/configuration.go
@@ -84,7 +84,6 @@ func NewConfiguration() *Configuration {
 		Debug:         false,
 		Servers: ServerConfigurations{
 			{
-				// Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code.
 				URL:         core.DefaultCloudURL,
 				Description: "No description provided",
 			},
--- a/tools/config/go-templates/configuration.mustache
+++ b/tools/config/go-templates/configuration.mustache
@@ -111,7 +111,9 @@ func NewConfiguration() *Configuration {
 		Servers:          ServerConfigurations{
 		{{/-first}}
 			{
-				// Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code.
+				{{! X-XGEN-MODIFIED }}
+				{{! Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code. }}
+				{{! Assumes a single root server in the spec, enforced by validation in the transformer. }}
 				URL: core.DefaultCloudURL,
 				Description: "{{{description}}}{{^description}}No description provided{{/description}}",
 				{{#variables}}
--- a/tools/config/go-templates/configuration.mustache
+++ b/tools/config/go-templates/configuration.mustache
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+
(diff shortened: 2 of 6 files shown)


tools/config/go-templates/configuration.mustache
```diff
@@ -109,7 +111,8 @@ func NewConfiguration() *Configuration {
 		Servers:          ServerConfigurations{
 		{{/-first}}
 			{
-				URL: "{{{url}}}",
+				// Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code.
```
AgustinBettati, line 114: Worth double checking: 
tools/CONTRIBUTING.md asks for {{! X-XGEN-MODIFIED }} on in-place template edits. This file has no X-XGEN markers today, so it is still a pristine upstream copy. Without the marker this edit is easy to lose on the next resync from openapi-generator.

admin/configuration.go
```diff
@@ -82,7 +84,8 @@ func NewConfiguration() *Configuration {
 		Debug:         false,
 		Servers: ServerConfigurations{
 			{
-				URL:         "https://cloud.mongodb.com",
+				// Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code.
```
AgustinBettati, line 87: nit: This renders into the public generated file and repeats once per server entry. A mustache comment ({{! ... }}) keeps the rationale in the template without shipping generation internals to consumers.

tools/config/go-templates/configuration.mustache
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+
```
AgustinBettati, line 9: Consistency only, nothing broken. OperationServers here still emits {{{url}}}, and Description at line 116 still comes from the spec, so "decoupled from the spec servers block" is not quite complete. Both are inert today: the spec has exactly one servers: key, at the root, with no operation-level blocks and no description.

AgustinBettati (approved):
LGTM, leaving some comments for extra consistency

tools/config/go-templates/configuration.mustache
```diff
@@ -109,7 +111,8 @@ func NewConfiguration() *Configuration {
 		Servers:          ServerConfigurations{
 		{{/-first}}
 			{
-				URL: "{{{url}}}",
+				// Default base URL is decoupled from the spec servers URL so spec changes do not affect generated code.
+				URL: core.DefaultCloudURL,
```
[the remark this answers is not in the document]
AgustinBettati: > If the spec ever defines multiple servers

Maybe we can add an assertion for this? So if this assumption ever breaking, we have clear guidelines on considerations to take.

tools/config/go-templates/configuration.mustache
oarbusi: Fixed, added {{! X-XGEN-MODIFIED }} with the rationale as mustache comments in [b183c7d](https://github.com/mongodb/atlas-sdk-go/pull/838/commits/b183c7d0d6c1b4d86a8dee3a15128bd664e4a483)

tools/config/go-templates/configuration.mustache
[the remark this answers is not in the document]
oarbusi: Added in [b183c7d](https://github.com/mongodb/atlas-sdk-go/pull/838/commits/b183c7d0d6c1b4d86a8dee3a15128bd664e4a483). Throws error on missing/multiple root servers, with guidance pointing to configuration.mustache. Unit tests added, and verified manually that a two-server spec fails generation.

admin/configuration.go
oarbusi: Done, moved to a mustache comment. The generated file now only shows `URL: core.DefaultCloudURL`

PR #842, opened 2026-08-17
feat: Add helper to detect ai agents

manupedrozo:
## Description

Adds a helper under a new `detectaiagent` package that identifies AI agents that invoke SDK consumers based on known markers (env vars & files).

Link to any related issue(s): CLOUDP-432500

## Type of change:

## Required Checklist:

### DIFF
--- a/detectaiagent/detect.go
+++ b/detectaiagent/detect.go
@@ -1,0 +1,151 @@
+// Package detectaiagent identifies AI agents that invoke SDK consumers.
+// Detection is based on known environment and filesystem markers used by AI agents.
+package detectaiagent
+
+import "os"
+
+const userAgentPrefix = "ai-agent/"
+
+// DetectedAgent describes an AI agent found by Detect.
+type DetectedAgent struct {
+	// ID is the agent identifier.
+	ID string
+	// UserAgentIdentifier is the agent identifier formatted for a User-Agent header (ai-agent/<ID>).
+	UserAgentIdentifier string
+}
+
+type agentDefinition struct {
+	id      string
+	markers []marker
+}
+
+type marker struct {
+	envVar   string
+	value    string
+	filePath string
+}
+
+var agentDefinitions = []agentDefinition{
+	{
+		id: "cursor",
+		markers: []marker{
+			{envVar: "CURSOR_AGENT"},
+			{envVar: "CURSOR_EXTENSION_HOST_ROLE", value: "agent-exec"},
+		},
+	},
+	{
+		id: "kimi",
+		markers: []marker{
+			{envVar: "KIMI_PLUGIN_ROOT"},
+		},
+	},
+	{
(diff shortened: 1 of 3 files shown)


detectaiagent/detect.go
```diff
@@ -0,0 +1,151 @@
...
+	},
+	{
+		id: "trae_ai",
+		markers: []marker{
+			{envVar: "TRAE_AI_SHELL_ID"},
+		},
+	},
+}
+
+// Detect returns the first AI agent detected from known markers.
```
oarbusi, line 127: This returns the first match in agentDefinitions, so when AI tools are composed, where one agent programmatically invokes another agent (e.g. claude code acting as the orchestrator and codex as the implementer), only one of them is ever reported, and which one wins is determined by list ordering rather than by anything meaningful.

I think we should detect all matches and return all so we properly capture the nested executions.

manupedrozo: Yes, this was a conscious decision made during design. We detect only the first one following the marker order.
Similar to CLI implementation (which will be replaced with this one): https://github.com/mongodb/mongodb-atlas-cli/blob/b064fbca010796273503e4d3e0ca51252d4fb73c/internal/telemetry/event.go#L154C5-L188

oarbusi: Any strong reason for just getting the first one instead of detecting all of them? how do we decide the order of the `agentDefinitions` which influences the agent we detect first in nested execution cases?  it feels like an easy win to detect all

manupedrozo: We could expand to multiple detection in the future but as of now we just want to report one via user-agent headers and telemetry, the main info we are looking for is whether tools are being used through agents, which particular agent is less relevant. So reporting multiple would introduce complexity on the analytics side at little benefit.
Note that similar tools such as https://github.com/vercel/detect-agent behave similarly.

oarbusi: Shipping first-match-only now and expanding later has a hidden cost: every tool needs to adopt the new SDK version before nested cases show up in analytics, and that adoption takes time. The change is small, so doing it now feels cheaper than later.
On the particular agent being less relevant: if that's true, we could just report ai-agent with no ID. But we do capture the specific agent, so it clearly matters. And in nested runs we lose that info for every agent but one. I tested this with opencode orchestrating claude: both OPENCODE and CLAUDECODE are in the consumer's env, but only one is reported, decided purely by slice ordering.
On vercel/detect-agent: it's a few months old in a brand new space, I wouldn't read it as best practice yet.

Not a blocker from my side, just think it's worth reconsidering.

manupedrozo: Thanks for the push back, I confirmed again with Victor (owner from the Analytics team) that we want first match only as agreed in the document.

AgustinBettati (approved):
LGTM

PR #850, opened 2026-08-31
fix: Reject dot-segment path parameters in request paths

oarbusi:
## Summary

Path parameters equal to `.` or `..` are now rejected client-side before any request is sent. `url.PathEscape` does not escape dot-segments, so such values would be sent as live path segments; servers normalize them per RFC 3986 before routing, which can route the request to a different endpoint than intended.

Changes are limited to the generation templates:

- `tools/config/go-templates/api.mustache`: emit a guard rejecting `.`/`..` values next to the existing empty check for every path parameter.
- `tools/config/go-templates/client.mustache`: defense-in-depth check in `prepareRequest` that refuses to send any assembled path containing a dot-segment. The check runs on the encoded path, so escaped values like `a%2F..%2Fb` remain valid.

The generated `admin/` changes are intentionally **not** included in this PR; the SDK autogeneration action will produce them from these templates after merge, avoiding conflicts with the auto-release flow.

## Tests

- Adds `test/path_traversal_test.go` covering: rejection of `.`/`..` parameters (asserting no request reaches the network), that non-dot-segment values (including `a/../b`, escaped to `a%2F..%2Fb`) are unaffected, and the `prepareRequest` guard including `%2E%2E`.
- Note: the Go test job will fail on this PR until the autogeneration action regenerates `admin/` with the new guards. The tests were run successfully (including `-race`) with the SDK regenerated locally from these templates, and the behavior was additionally verified with read-only live calls against a real Atlas instance.

Ticket: [SECBUG-4400](https://jira.mongodb.org/browse/SECBUG-4400)

### DIFF
--- a/test/path_traversal_test.go
+++ b/test/path_traversal_test.go
@@ -1,0 +1,124 @@
+package test
+
+// Path parameters equal to dot-segments ("." or "..") must be rejected
+// client-side before any request is sent: the server normalizes them per
+// RFC 3986, which would route the authenticated request to a different
+// endpoint than intended.
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.mongodb.org/atlas-sdk/v20250312023/admin"
+)
+
+func TestDotSegmentPathParamsRejected(t *testing.T) {
+	var hits int32
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		atomic.AddInt32(&hits, 1)
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("{}"))
+	}))
+	defer server.Close()
+
+	sdk, err := admin.NewClient(admin.UseBaseURL(server.URL))
+	require.NoError(t, err)
+	ctx := context.Background()
+
+	cases := []struct {
+		name      string
+		paramName string
+		call      func() error
+	}{
+		{"DeleteIndexByName groupId=..", "groupId", func() error {
+			_, err := sdk.AtlasSearchAPI.DeleteIndexByName(ctx, "..", "cluster0", "coll", "db", "idx").Execute()
+			return err
+		}},
+		{"DeleteIndexByName collectionName=..", "collectionName", func() error {
+			_, err := sdk.AtlasSearchAPI.DeleteIndexByName(ctx, "group1", "cluster0", "..", "db", "idx").Execute()
+			return err
+		}},
+		{"DeleteIndexByName collectionName=.", "collectionName", func() error {
+			_, err := sdk.AtlasSearchAPI.DeleteIndexByName(ctx, "group1", "cluster0", ".", "db", "idx").Execute()
+			return err
+		}},
+		{"GetDatabaseUser username=..", "username", func() error {
+			_, _, err := sdk.DatabaseUsersAPI.GetDatabaseUser(ctx, "group1", "admin", "..").Execute()
+			return err
+		}},
+	}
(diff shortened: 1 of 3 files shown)


tools/config/go-templates/api.mustache
```diff
@@ -157,6 +157,9 @@ func (a *{{{classname}}}Service) {{{nickname}}}Execute(r {{operationId}}ApiReque
 	if r.{{paramName}} == "" {
 		return {{#returnType}}localVarReturnValue, {{/returnType}}nil, reportError("{{paramName}} is empty and must be specified")
 	}
+	if r.{{paramName}} == "." || r.{{paramName}} == ".." {
```
EspenAlbert, line 160: q: Should this also cover ../.. or other similar patterns or are those already handled by the `prepareRequest` maybe?

EspenAlbert (approved):
TY!
*Looks like CI is failing

oarbusi:
@EspenAlbert yes CI is failing because I didn't include the autogeneration in this PR so that we do a clean autogen once this is merged, and the new test fail due to not having the new changes autogenerated. I have run test locally and they are all successful

test/path_traversal_test.go
```diff
@@ -0,0 +1,133 @@
...
+			before := atomic.LoadInt32(&hits)
+			err := tc.call()
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), tc.paramName+" must not be a dot-segment path parameter")
+			assert.Equal(t, before, atomic.LoadInt32(&hits), "rejected call must not reach the network")
+		})
+	}
+}
+
+func TestNonDotSegmentPathParamsStillWork(t *testing.T) {
```
manupedrozo, line 77: Nice!

tools/config/go-templates/api.mustache
oarbusi: No need. The SDK escapes `/` to `%2F` before sending, so `../..` is sent as one harmless segment (`..%2F..`). Only a value that is exactly `.` or `..` can cause trouble, and this guard blocks those. There's a test for this: `a/../b` goes out as `a%2F..%2Fb`

test/path_traversal_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NoError(t, err)
+	ctx := context.Background()
+
+	cases := []struct {
+		name       string
+		collection string
+		wantInURI  string
+	}{
+		{"dotted name", "my.coll", "/my.coll/"},
+		{"three dots is not a dot-segment", "...", "/.../"},
```
manupedrozo, line 97: Just double checking: are we sure that 3+ dots are ok?

tools/config/go-templates/client.mustache
```diff
@@ -313,6 +313,18 @@ func (c *APIClient) prepareRequest(
 		return nil, err
 	}
 
+	// Defense-in-depth: refuse to send paths with dot-segments, which the
```
manupedrozo, line 316: Nice! 
This check **IS** necessary for for the untyped client since calls do not go through `Execute()`, correct? 
Worth adjusting the wording if so.

manupedrozo (approved):

test/path_traversal_test.go
oarbusi: Yes. Only exact `.` and `..` segments get normalized away by the server. `...` is just a normal name, it goes out untouched as `/.../`

tools/config/go-templates/client.mustache
oarbusi: Correct, `UntypedClient.PrepareRequest` takes a raw path straight to `prepareRequest`, skipping the generated guards, so this check is the only protection there. Reworded the comment in [7333b8b](https://github.com/mongodb/atlas-sdk-go/pull/850/commits/7333b8bedcd33a3c4724cedf93960b41823e74dc)

test/path_traversal_test.go
```diff
@@ -0,0 +1,133 @@
...
+		{"dots inside a name", "a..b", "/a..b/"},
```
AgustinBettati: nice

PR #856, opened 2026-09-14
chore: Bump Go to 1.26 in the SDK modules

EspenAlbert:
## Description

Raises the `go` directive from `1.25.0` to `1.26.0` in the SDK modules:

- `go.mod`
- `examples/go.mod`

The `examples` module was the one left behind. Its `replace` points at the root module, so it cannot resolve a parent that requires 1.26 while it declares 1.25. This is what makes [dependabot PR #855](https://github.com/mongodb/atlas-sdk-go/pull/855) fail today:

```
go: updates to go.mod needed; to update it:
	go mod tidy
make: *** [Makefile:24: test-examples] Error 1
```

No dependency versions change here. `golang.org/x/oauth2` stays at `v0.36.0`.

## Why this is separate from the dependency bump

PR #855 bumps `golang.org/x/oauth2` from `0.36.0` to `0.37.0`. Since Go 1.21 the `go` directive is a hard minimum rather than advisory, and `v0.37.0` declares `go 1.26.0` in its own `go.mod`, so any user of that release must be on Go 1.26. Dependabot shells out to `go get`, which is why the version change landed inside a `chore(deps)` PR.

Splitting it out keeps the dependency PR to a dependency diff and records the version bump on its own, matching `9f01cf7f` (Go 1.25) and `7401f2e4`.

PR #855 should be rebased after this merges. Its diff then reduces to the `oauth2` requirement and `go.sum`.

## Notes for reviewers

- No `toolchain` directive is added. It does not prevent the `go` directive from moving, so it would only add noise.
- `golangci-lint` stays at `v2.11.4`. Go 1.26 has been supported since `v2.9.0` and the existing pin already lints this successfully.
- Downstream modules are already at or above this floor: `terraform-provider-mongodbatlas` on `go 1.26.4`, `cfn-resources` on `go 1.26.2`, `cli-plugin-tf` on `go 1.26.0`.
- `CONTRIBUTING.md` stated "at least Go 1.21" and claimed the repo uses asdf for Go version management. `.tool-versions` only pins Node, so the stale statement was replaced with a pointer to the `go` directive in `go.mod`, and the asdf line was removed. The doc no longer needs editing on each version bump.

## Type of change:

- [x] This change requires a documentation update

## Required Checklist:


marcabreracast (approved):
LGTM!

PR #857, opened 2026-09-14
chore: Updates both Go modules in one Dependabot PR

EspenAlbert:
## Description

Two changes that fix the same failure.

**Dependabot now updates both Go modules together.** The config listed `/` and `/examples` as separate gomod entries on different schedules, weekly and monthly. The examples module depends on the root module through a local `replace`, so a dependency bump that touches only the root leaves the examples module stale and CI fails. The entries are merged into one `directories` entry with `group-by: dependency-name`, which raises each dependency in both modules in a single PR.

**Lands `golang.org/x/oauth2` v0.37.0 in both modules.** This is the bump from [dependabot PR #855](https://github.com/mongodb/atlas-sdk-go/pull/855), plus the examples half that PR was missing. It supersedes #855.

Related issue(s): supersedes [dependabot PR #855](https://github.com/mongodb/atlas-sdk-go/pull/855)

## Why #855 was failing after the Go bump

#855 raised `go.mod` to `oauth2 v0.37.0` but left `examples/go.mod` on `v0.36.0`. The examples module carries its own indirect pin, because it depends on the SDK through `replace go.mongodb.org/atlas-sdk/v20250312025 => ../`. Minimum version selection picks `v0.37.0` from the parent, so the examples module file no longer matched its own graph and `go test` refused to run:

```
go: downloading golang.org/x/oauth2 v0.36.0
go: updates to go.mod needed; to update it:
	go mod tidy
make: *** [Makefile:24: test-examples] Error 1
```

Bumping Go to 1.26 in #856 fixed the `go` directive mismatch but not this one. Both modules need the same dependency version.

## Notes for reviewers

- `group-by` is a child of the group entry, not of the update entry. The comment in `dependabot.yml` explains in one sentence why the entries are merged.
- The merged entry keeps the weekly schedule. The examples entry ran monthly, which is part of why the modules drifted apart.
- The `github-actions` and `npm` entries are unchanged.
- `group-by: dependency-name` requires all directories to use the same ecosystem, applies to version updates only, and falls back to separate PRs when constraints differ across directories.

## Type of change:

## Required Checklist:


.github/dependabot.yml
```diff
@@ -11,17 +14,9 @@ updates:
...
-      day: sunday
-    allow:
-      - dependency-type: "all"
-    groups:
-      gomod:
-        patterns:
-          - "*"
+        # Without this, `directories` still opens one PR per directory, so a bump could land in
+        # `/` while `/examples` keeps the old pin and `make test-examples` fails.
+        group-by: dependency-name
```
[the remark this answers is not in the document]
EspenAlbert: /ar existing Dependabot policy: this PR preserves the weekly schedule and existing batch scope; it only merges the root and examples gomod entries so the linked module graph stays synchronized

.github/dependabot.yml
```diff
@@ -1,32 +1,33 @@
...
+    directories:
+      - "/"
+      - "/examples"
     schedule:
       interval: weekly
       day: sunday
+    # Wait 7 days on new releases before proposing them, so a compromised or broken
+    # version has time to be caught. Security updates are never delayed. Set on every entry.
+    cooldown:
+      default-days: 7
```
manupedrozo: Seems like the default cooldown is 3: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-
What is the reason for having a higher one? Should we apply it across all our repos?

EspenAlbert: Good spot. 
The 7-day value came from Semgrep's suggested remediation rather than a repository policy. This task is limited to synchronizing the two gomod modules and explicitly leaves the GitHub Actions and npm entries unchanged, so I’ll remove the cooldown changes; a cross-repository policy needs a separate decision

manupedrozo: 🤖

manupedrozo (approved):

